PR #102, opened 2021-04-27
[PCX-1472] Display main UPDATE page

Sorix:
**AS AN** end-customer
**I WANT** to enter the UPDATE flow main page
**SO THAT** I can potentially edit or delete my saved payment methods or add new ones

This is main page that is loaded when initially entering the UPDATE flow, and every time after an update `operation` (`/update`  or `/delete` endpoints).
This page has two variants:
1. The _account variant_: displays only saved methods (`accounts`)
2. A _network variant_: displays only available networks (`networks`)

While not directly in the scope of this ticket, take note (in terms of implementation) that while initially the page may load in either of two variants (depending on the and {{account}}s and {{network}}s in the LIST object), a different variant may need to load after adding/removing/updating payment methods. For example, the initial LIST object contains one saved method, therefore variant 1 is displayed, the method is deleted, therefore variant 2 needs to be loaded. 
### Acceptance criteria
1. All texts (keys in parentheses) are mocked, in English only, until such time that they are added to the language files and translated properly
1. The spinner is shown until the correct variant is fully displayed
1. The UPDATE main page includes:
    1. A “back” button
    2. If saved accounts exist in the UPDATE LIST object:
        1. The title “Payment methods“ (`paymentpage.title`)
        2. The sub-title “Edit your payment methods” (`accounts.UPDATE.title`)
        3. The saved methods themselves, the same way they’re displayed on the payment flow
    1. If no saved accounts exist:
        1. The title “Add a payment method“ (`addmethodpage.title`)
        2. All networks just as they would be displayed on the payment flow
1. Design follows [PCX-1388: Android SDK UPDATE flow](https://optile.atlassian.net/browse/PCX-1388)


Sorix:
❌ **Ticket was cancelled.**

Replaced by #106.

PR #111, opened 2021-05-30
[PCX-1385] Delete saved payment method

Sorix:

PR #113, opened 2021-06-21
[PCX-1701] Hide "Update" button for registered methods with empty forms

Sorix:
AS AN end-customer
I WANT TO not see an “Update” button registered method bottom sheets if the method has no form
SO THAT I don’t receive an error when clicking it

### Acceptance criteria
1. Bottom sheets for registered methods with an empty form only display the account mask and the trashcan icon if they have an empty form
2. The above does not affect the PAYMENT flow

_That PR includes [PCX-1385](https://github.com/optile/checkout-ios/pull/111), review only after accepting it._

### DIFF
--- a/Sources/UI/Scenes/Input/Model/Input.Network.swift
+++ b/Sources/UI/Scenes/Input/Model/Input.Network.swift
@@ -61,7 +61,7 @@ extension Input.Network {
 
         let submitButton: Input.Field.Button?
         init(networkLabel: String, maskedAccountLabel: String?, logo: UIImage?, inputFields: [InputField], separatedCheckboxes: [InputField], submitButton: Input.Field.Button?) {
-                self.networkLabel = networkLabel
+            self.networkLabel = networkLabel
             self.maskedAccountLabel = maskedAccountLabel
             self.logo = logo
             self.inputFields = inputFields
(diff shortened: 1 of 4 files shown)


Sources/UI/Scenes/Input/Model/Input.Network.swift
```diff
@@ -59,9 +59,9 @@ extension Input.Network {
         /// Checkboxes that must be arranged in another section (used for recurrence and registration)
         let separatedCheckboxes: [InputField]
 
-        let submitButton: Input.Field.Button
-        init(networkLabel: String, maskedAccountLabel: String?, logo: UIImage?, inputFields: [InputField], separatedCheckboxes: [InputField], submitButton: Input.Field.Button) {
-            self.networkLabel = networkLabel
+        let submitButton: Input.Field.Button?
+        init(networkLabel: String, maskedAccountLabel: String?, logo: UIImage?, inputFields: [InputField], separatedCheckboxes: [InputField], submitButton: Input.Field.Button?) {
+                self.networkLabel = networkLabel
```
siebeprojects, line 64: There are 4 characters too much I think.

siebeprojects (approved):

PR #115, opened 2021-07-05
[PCX-1396] Add and save new payment method

Sorix:
**AS A** customer
**I WANT** to save a new payment method
**SO THAT** I can later use it to make purchases

### Acceptance criteria
1. The available `networks` are in essence displayed the same way they are displayed in the PAYMENT flow. An end-customer is able to create a new registration using the methods listed without making a payment
    1. The one exception to the above is that `networks` with a merchant registration configuration of `NONE` /`NONE` are hidden (a bug in OPG returns them, will be fixed in the scope of https://optile.atlassian.net/browse/PNC-255 but for now we have to filter them out)
1. The action button uses the text based on the LIST object’s `operation` - in this case `button.operation.UPDATE.label` (“Save”)
1. No registration options are displayed (in contrast to the PAYMENT flow)
1. Clicking the “Save” button triggers the `operation` URL the same way it would be triggered in the PAYMENT flow, except that the endpoint is now `/update`. 
    1. `autoRegistration` and `allowRecurrence` are determined based entirely on the merchant configuration flags for the method (see table below)
    1. A spinner (and dimmed background) is displayed until a response is received (as in the PAYMENT flow)
1. The response from the `/update` call is handled as per the interaction code/reason:
    1. `RELOAD`/`*`:
        1. Reload the LIST object; re-render the activity
    1. `PROCEED`/`PENDING`:
        1. A message is shown to the customer indicating that the method saving is in process and might not appear right away. We hard-code the English message from [PCX-1661: Indicate a pending registration on UPDATE flow](https://optile.atlassian.net/browse/PCX-1661)
    1. `TRY_OTHER_ACCOUNT`/`*` or `TRY_OTHER_NETWORK`/`*`, `RETRY`/`*`:
        1. The LIST object is reloaded; the scene is re-rendered
        1. A popup with the default `interaction.*` title and text is shown, with a single dismissal button (`button.OK.label`)
     1. `ABORT`/`*` (list doesn’t exist = `ABORT`/`SYSTEM_FAILURE`) we return control to the merchant with the `paymentResult`
    1. For any interaction code not listed in a-d or `PROCEED` with an interaction reason different than `PENDING`, treat it as an `ABORT`


.circleci/config.yml
```diff
@@ -91,10 +91,10 @@ jobs:
           path: output/scan
       - run:
           name: Zip xcresult
-          command: zip -r output/scan/example.xcresult.zip output/scan/Example.xcresult/
+          command: zip -r output/scan/checkout.xcresult.zip output/scan/ExampleCheckout.xcresult/
```
siebeprojects, line 94: why not call it examplecheckout.xcresult.zip instead of checkout.xcresult.zip?

ExampleCheckout/UITests/Assets/Transaction.json
```diff
@@ -1,11 +1,13 @@
 {
   "transactionId": "tr1",
   "integration": "MOBILE_NATIVE",
-  "presetFirst": false,
   "country": "DE",
   "customer": {
     "number": "1",
-    "email": "john.doe@example.com"
+    "email": "john.doe@example.com",
+    "registration": {
```
siebeprojects, line 8: In the checkout android SDK the tests are done without the registration id. So, in order to test delete the sequence is: 1. register new account, 2. open the page, 3 click delete button and then check if the newly created account is gone again.
Using the registration ID is not a blocker but it may cause problems later when the integration server is rebuild and we loose the registration ID.

Sources/Localization/Strings/TranslationKey.swift
```diff
@@ -17,6 +17,10 @@ enum TranslationKey: String, CaseIterable {
     case accountsUpdateTitle = "accounts.update.title"
     case networksUpdateTitle = "networks.update.title"
 
+    // Used in UPDATE flow
```
siebeprojects, line 20: How is it determined that this is only used in UPDATE flow? 
Is that because we do not show a popup dialog for PROCEED/PENDING in CHARGE flow? 
If we would show a popup for PROCEED/PENDING in CHARGE flow, would it then show the same title/message as defined here?

ExampleCheckout/UITests/Assets/Transaction.json
Sorix: >  it may cause problems later when the integration server is rebuild and we loose the registration ID
It will cause problems in "Delete button shouldn't appear in CHARGE flow", because I save the new payment method using UPDATE flow, after that, I need to regenerate the List URL in a CHARGE flow to go to that saved method and check if delete button actually appeared.

I don't know how to get registration id from a server if we don't specify it in `Transaction` when I create the payment session in UPDATE flow, at least in UI tests (they don't have access to servers replies).

.circleci/config.yml
Sorix: I just don't want to use "Example" as a test-report filename. We're running UI tests in Example application, but it is a current use-case, that may change in the future and I don't want to connect UI tests with "Example" naming, they're not testing Example app, they're testing our framework, that application is used only like a "proxy" (that may change in future).

siebeprojects (approved):

PR #120, opened 2021-08-09
[PCX-1652] Update to iOS14 Logger system

Sorix:
**AS A** developer
**I WANT TO** to see readable logs in output
**SO THAT** logging system should be refined

### Acceptance Criteria
1. Logs are beatifully formatted with correct Input & Output indication.
2. The iOS14 logging system is supported.
3. Private data are hidden or redacted on release builds.

### DIFF
--- a/Sources/UI/Scenes/Input/Input.ViewController.swift
+++ b/Sources/UI/Scenes/Input/Input.ViewController.swift
@@ -7,6 +7,7 @@
 #if canImport(UIKit)
 import UIKit
 import SafariServices
+import os.log
 
 // MARK: Initializers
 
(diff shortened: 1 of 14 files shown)


Sources/UI/Scenes/Input/Input.ViewController.swift
```diff
@@ -242,8 +246,18 @@ extension Input.ViewController: InputTableControllerDelegate {
     private func replaceCurrentNetwork(with newSelection: Input.SmartSwitch.Selector.DetectedNetwork) {
         if let imagesHeaderModel = header as? Input.ImagesHeader {
             switch newSelection {
-            case .generic: imagesHeaderModel.networks = self.networks
-            case .specific(let specificNetwork): imagesHeaderModel.networks = [specificNetwork]
+            case .generic:
+                if #available(iOS 14.0, *) {
```
siebeprojects, line 250: Could the #available check for version 14 be unified and done in a separate Logger class? 
So, the logger class would do the check instead of having this check each time we want to log.

Sorix: Unfortunately, I can't do that because I need to create `OSLogMessage` (that's an argument for a logger with text and variables) in one place, I can't "redirect" it or pass it as a parameter.

> **Apple Documentation**
> Always specify a string; don’t create an `OSLogMessage` structure and pass it in as a parameter.

Sorix: We could plan a call, I will describe we I dropped logging for the previous version of iOS.

siebeprojects (approved):

PR #123, opened 2021-08-20
[PCX-1935] Fix OSLog issues

caiomello:
- Replaced instances of `import OSLog` with `import os.log` ([More info](https://developer.apple.com/forums/thread/661061))
- Replaced calls to `os_log` with instances of `Logger`


.gitignore
```diff
@@ -21,6 +21,10 @@ DerivedData/
 !default.mode2v3
 *.perspectivev3
 !default.perspectivev3
+.DS_Store
```
Sorix, line 24: I prefer to keep OS-related files in [global gitignore](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) rather than in repo-specific. @siebeprojects what do you think about that?

Sources/Core/log.swift
```diff
@@ -5,10 +5,13 @@
 // See the LICENSE file for more information.
 
 import Foundation
-import os
+import os.log
 
 func log(_ type: LogType, _ message: StaticString, _ args: CVarArg...) {
```
Sorix, line 10: I think we should get rid of duplication: `InternalError().log` and global `log()`. Let's keep only one function (I think global `log()` should be eliminated but it's up to you to decide).

Sources/Core/log.swift
```diff
@@ -5,10 +5,13 @@
...
-    os_log(type.osLogType, message, args)
+    if #available(iOS 14.0, *) {
+        let logger = Logger(subsystem: Bundle.frameworkIdentifier + ".network", category: "Log")
```
Sorix, line 12: That's not a network subsystem (it's a global function). If you decide to remove that function, that comment doesn't matter in another case we should remove it.

```suggestion
        let logger = Logger(subsystem: Bundle.frameworkIdentifier, category: "Log")
```

Sorix (changes requested):

.gitignore
caiomello: I wasn't aware of this, good to know! Still, any other/future dev that doesn't have a global gitignore will run into this, and I don't see any reason not to have this in the local ignore anyway... I'll revert the change.

Sources/Core/log.swift
caiomello: Now that I've looked into the code some more I think there are too many different mechanisms being used for logging. Sadly we cannot wrap `Logger` under an abstraction, so calls to `Logger` will always be pervasive throughout the codebase. What we can do instead is have a unified approach, where any usage of `Logger` follows the same pattern. Right now we have multiple helper `log` functions in different places, like in `log.swift`, `InternalError`, various local/public extensions, and as part of the `Request` protocol. It'd be great to streamline this, for example by enforcing that there must always be a local `logger` property that the object can use, like it's already implemented in the input and list view controllers.

Sorix: @caiomello please check 6874051ab8e0f8279fd823a0d304793050d8e893

caiomello: I was trying some stuff out yesterday and started with something like this:

image

The protocol approach is good, I like it.

Sorix (approved):

siebeprojects (approved):

Sorix:
@caiomello please update from `develop`

PR #124, opened 2021-08-23
[PCX-1809] Support flow-specific translations

Sorix:
**AS AN** end-customer
**I WANT** to receive meaningful error or informational messages in response to my actions 
**SO THAT** I understand what went wrong or why the application behaves in a certain way

> Definition: Let the flow be defined as the `operationType` in the root of the LIST object if the OPG response (interaction code/reason) was triggered by clicking the action button (e.g. “Pay”, “Save”, “Credit”; in other words the `operationURL` endpoint was called), and “DELETE” if the OPG response was triggered by deleting a payment method, regardless of the `operationType`.
>
> In other words, flow can be either `CHARGE`, `UPDATE`, `PAYOUT` (or any other `operationType`) or `DELETE` (so-called “virtual” flow, since it does not map to an `operationType`, because deleting a payment method can be carried out regardless of `operationType`. 
>
> In this ticket we will enable the iOS SDK to load different error/info messages in response to the same interaction codes/reasons based on the flow. For more information read Suggestion 2 on https://optile.atlassian.net/wiki/spaces/PPW/pages/2779709539/Flow+specific+translations)

### Acceptance criteria
1. All interaction messages (`interaction.*` keys) from the `checkout` properties file are loaded first in their flow-specific version (if exists): `interaction.<flow>.<interactionCode>.<interactionReason>.*` (“*” can be “text”,  “title”, etc.) based on the flow definition above, which means it depends on the customer interaction that resulted in the OPG response (interaction code/reason):
  1. If the interaction triggered a call to an `operationURL` end-point, the flow is the `operationType`
  1. If the interaction triggered a `DELETE` call on a registered payment method, the flow is a `DELETE`
1. If the flow-specific version does not exist, the default interaction message `interaction.<interactionCode>.<interactionReason>.*` is used (“Fallback”)


Sources/UI/Scenes/Input/Input.ViewController.OperationResultHandler.swift
```diff
@@ -24,6 +24,12 @@ protocol InputPaymentControllerDelegate: AnyObject {
 extension Input.ViewController {
     class OperationResultHandler {
         weak var delegate: InputPaymentControllerDelegate?
+
+        let listOperationType: String
```
caiomello, line 28: I may lack some context here, but does this need to be a `String`? Why is it not an `enum`?

siebeprojects (approved):

caiomello (approved):

PR #125, opened 2021-08-27
[PCX-1930] Create Markdown parser

caiomello:


Tests/Core/MarkdownParserTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    func testParseLinks_whenInputContains2LinksWithoutWhitespace_shouldReturnAttributedStringWith2Links() {
+        let input = "By clicking the button, you agree to the [Terms of Service](https://www.apple.com/)and[Privacy Policy](https://www.google.com/)."
+        let links = parser.parseLinks(in: input)
+        XCTAssertEqual(links.count, 2)
+    }
+
+    func testParseLinks_whenInvalidURL_shouldReturn0Links() {
+        let input = "By clicking the button, you agree to the [Terms of Service]()."
```
Sorix, line 74: What will be the result of parsing that input?

caiomello: The same string, `By clicking the button, you agree to the [Terms of Service]().`

Sources/Core/MarkdownParser.swift
```diff
@@ -0,0 +1,86 @@
...
+private let textExpression = "\\[(?<text>[^\\]]*)\\]"
+private let urlExpression = "\\((?<url>[^\\)]*)\\)"
+private let linkExpression = textExpression + urlExpression
+
+/// Responsible for parsing Markdown. Currently only supports links.
+struct MarkdownParser: Loggable {
+    /// Parses a Markdown string into an `NSAttributedString`.
+    /// - Parameter input: The string to be parsed.
+    /// - Returns: An `NSAttributedString` containing the parsed Markdown elements.
+    func parse(_ input: String) -> NSAttributedString {
```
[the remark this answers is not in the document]
caiomello: I didn't think a label would add any clarity in this case. When you call `parse` on a Markdown parser, it's clear that you're giving it some Markdown text. The function already tells you the argument is a `String`, and it doesn't even need to be Markdown, it's 100% fine to give it any `String`, it'll just convert it to an `NSAttributedString` with the same text if it doesn't contain any links.

Sorix (approved):

Tests/Core/MarkdownParserTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    func testParseLinks_whenInputContains2LinksTogether_shouldReturnAttributedStringWith2Links() {
+        let input = "By clicking the button, you agree to the [Terms of Service](https://www.apple.com/)[Privacy Policy](https://www.google.com/)."
+        let links = parser.parseLinks(in: input)
+        XCTAssertEqual(links.count, 2)
+    }
+
+    func testParseLinks_whenInputContains2LinksWithoutWhitespace_shouldReturnAttributedStringWith2Links() {
+        let input = "By clicking the button, you agree to the [Terms of Service](https://www.apple.com/)and[Privacy Policy](https://www.google.com/)."
```
bzlat, line 68: What would happen if I also add a `title` to the `link`?
```
A link with title: [Link Text](https://www.google.com "link title")
```

caiomello: It's not a valid URL, so it would return the same string:
```
A link with a title: [Link Text](https://www.google.com "link title")
```

Is that valid Markdown?

caiomello: Solved in https://github.com/optile/checkout-ios/pull/125/commits/6405a8c0e6504bec6767b900bee1e53d480f663c.

bzlat (approved):

Sorix:
@caiomello, please merge

Sources/Core/MarkdownParser.swift
```diff
@@ -0,0 +1,91 @@
...
+
+            // Parse text: Link
+            guard let textRange = string.range(of: textExpression, options: .regularExpression) else { return nil }
+            let text = string[textRange].dropFirst().dropLast()
+
+            // Parse URL: <url>
+            // Ignore optional title: "title"
+            guard let urlComponentRange = string.range(of: urlExpression, options: .regularExpression) else { return nil } // (<url> "title")
+            let urlComponentText = String(string[urlComponentRange].dropFirst().dropLast()) // <url> "title"
+            let urlString = urlComponentText.components(separatedBy: " ").first ?? urlComponentText // <url>
```
Sorix, line 78:
```suggestion
            let urlString = urlComponentText.components(separatedBy: .whitespaces).first ?? urlComponentText // <url>
```

Sorix (approved):

Tests/Core/MarkdownParserTests.swift
```diff
@@ -0,0 +1,90 @@
...
+    func testParse_whenInputContainsLinks_shouldReturnStringContainingLinkText() {
+        let input = "By clicking the button, you agree to the [Terms of Service](https://www.apple.com/) and [Privacy Policy](https://www.google.com/)."
+        let result = parser.parse(input)
+        XCTAssertEqual(result.string, "By clicking the button, you agree to the Terms of Service and Privacy Policy.")
+    }
+
+    func testParse_whenInputContainsLinkWithOptionalTitle_shouldReturnStringContainingLinkText() {
+        let input = "By clicking the button, you agree to the [Terms of Service](https://www.apple.com/ \"optional title\")."
+        let result = parser.parse(input)
+        XCTAssertEqual(result.string, "By clicking the button, you agree to the Terms of Service.")
```
siebeprojects, line 46: It would be good to also test that the parsed URL does not contain the "optional string". So, instead of only testing the end result, it should also test the contents of the URL and optional title.

PR #130, opened 2021-09-06
[PCX-1930] Allow custom text above/beneath method forms in PAYMENT flow

Sorix:
**AS A** merchant
**I WANT** to display my customers a text above or beneath method forms in the PAYMENT flow
**SO THAT** I can inform them of business specific practices, alert them what happens when they click “Pay”, direct them to terms and conditions, etc.

### Acceptance criteria
1. Texts appear either above payment method forms or below them (based on the `top`/`bottom` MCAPI configuration, see [Extra Elements (Previously Custom Checkboxes)](https://optile.atlassian.net/wiki/spaces/PPW/pages/2801795107/Custom+Checkboxes)) when applicable (when a form exists). If a form does not exist, the `top`/`bottom` configuration has no effect (configuring either of them has the same end result)
1. The texts can include one or multiple hyperlinked texts (any sequence of characters)
1. Links open in the default device browser

PR #132, opened 2021-09-07
[PCX-2012] Respect allowDelete on UPDATE flow

caiomello:


Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.swift
```diff
@@ -60,8 +60,7 @@ extension Input.ModelTransformer {
             throw InternalError(description: "Incorrect registered account model, operation URL is not present. Links: %@", objects: registeredAccount.apiModel.links)
         }
 
-        // Detect if we're in UPDATE flow
-        let isDeletable = registeredAccount.apiModel.operationType == "UPDATE"
+        let isDeletable = registeredAccount.isDeletable
```
Sorix, line 63: I think you could remove that line at all and just use `registeredAccount.isDeletable` here:

https://github.com/optile/checkout-ios/blob/45367564b15c38dfccfa6fc41829e013ee594d63/Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.swift#L89

Sources/UI/Service/PaymentSessionService/PaymentSessionProvider.swift
```diff
@@ -160,7 +160,11 @@ class PaymentSessionProvider {
             throw InternalError(description: "Operation type is not known or supported: %@", operationType)
         }
 
-        let context = PaymentContext(operationType: operation, extraElements: listResult?.extraElements)
+        let context = PaymentContext(
+            operationType: operation,
+            extraElements: listResult?.extraElements,
+            isDeletable: listResult?.allowDelete == false ? false : true // Defined in https://optile.atlassian.net/browse/PCX-2012
```
Sorix, line 166: It's hard readable (unwrapping + if in one line) , but it's up to you. I could just suggest to have a separate line for a comment to avoid too long line.

caiomello: 100% agree, I'll extract this value.

Sorix:
Waiting for #130 to merge, because `PaymentContext` was introduced there.

Sources/UI/Model/RegisteredAccount.swift
```diff
@@ -9,14 +9,16 @@ import UIKit
...
 
     let networkLabel: String
     let submitButtonLabel: String
     var logo: Loadable<UIImage>?
 
-    init(from apiModel: AccountRegistration, submitButtonLocalizationKey: String, localizeUsing localizer: TranslationProvider) {
+    init(from apiModel: AccountRegistration, submitButtonLocalizationKey: String, localizeUsing localizer: TranslationProvider, isDeletable: Bool) {
         self.apiModel = apiModel
         self.translation = localizer
+        self.isDeletable = isDeletable
```
Sorix, line 21: You decided to make that decision inside `PaymenSession`, now we could have inconsistency that decision about `PaymentNetwork` is done inside `ModelTransformer`. But `PaymentNetwork` couldn't be deleted by design of that entity...

https://github.com/optile/checkout-ios/blob/0037e7f1168fb21032fe1c843ef31a630cc94a27/Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.swift#L150

We have the next options:
1. Add `let isDeletable = false` in `PaymentNetwork`.
2. Make decisions in `ModelTransformer`.
2. Keep it like now (it's also ok).

I think because of that problem I decided some time ago to make that decision on `ModelTranformer` to avoid it. I don't know the best way to deal with that.

caiomello: Moved to `ModelTransformer` in https://github.com/optile/checkout-ios/pull/132/commits/32ffe1edb5b938700b93acab50a4c451dd686268.

Sorix (changes requested):
We have a UI test for the delete button that actually checks when that button should appear and when not.

UI tests doesn't support `allowDelete` flag, we should add support for it and test `CHARGE`, `UPDATE` flow with different flag states. There is a bunch of work for that :(.

https://github.com/optile/checkout-ios/blob/32ffe1edb5b938700b93acab50a4c451dd686268/ExampleCheckout/UITests/UpdateFlowTests.swift#L98-L126

Sorix:
@caiomello I think it will be better to create the new branch from `develop` and apply your changes to the new branch, than resolving that conflicts and finding the changes in code review that could accidentally lead to merging something unexpected.

caiomello:
> @caiomello I think it will be better to create the new branch from `develop` and apply your changes to the new branch, than resolving that conflicts and finding the changes in code review that could accidentally lead to merging something unexpected.

Very simple conflicts, no problem.

Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.swift
```diff
@@ -79,6 +76,18 @@ extension Input.ModelTransformer {
...
+        let isDeletable: Bool = {
+            switch registeredAccount.apiModel.operationType {
+            case "UPDATE":
+                return (registeredAccount.isDeletable == false) ? false : true
+            case "CHARGE":
+                return registeredAccount.isDeletable == true
+            default:
+                return false
+            }
+        }()
```
caiomello, line 89: @Sorix @siebeprojects Please verify.

Sorix: @caiomello Could you describe the purpose of `return (registeredAccount.isDeletable == false) ? false : true` Is registeredAccount still optional?

caiomello: `registeredAccount`'s `isDeletable` is optional.

`(registeredAccount.isDeletable == false) ? false : true` represents "The delete payment method (trashcan) is available if and only if allowDelete in the LIST object is either set to true or undefined in UPDATE flow". So the result is false only is `registeredAccount.isDeletable` is false.

Sources/UI/Service/PaymentSessionService/PaymentSessionProvider.swift
```diff
@@ -160,7 +160,7 @@ class PaymentSessionProvider {
             throw InternalError(description: "Operation type is not known or supported: %@", operationType)
         }
 
-        let context = PaymentContext(operationType: operation, extraElements: listResult?.extraElements)
+        let context = PaymentContext(operationType: operation, extraElements: listResult?.extraElements, isDeletable: listResult?.allowDelete)
```
Sorix, line 163: @caiomello I recommend avoiding increasing the count of `isDeletable` optionals in the code. I think it is better to set in the root to `true` or `false` to avoid those three-state Booleans in multiple places.

I think that logic should be located here, we could eliminate the backend's flaws with undefined Boolean and work with a precise state.

Sorix: I could propose doing something like that:

```swift
private func createPaymentSession(from translations: DownloadTranslationService.Translations) throws -> PaymentSession {
    guard let operationType = listResult?.operationType else {
        throw InternalError(description: "Operation type or ListResult is not defined")
    }

    // PaymentSession.Operation contains only supported operation types by the framework
    guard let operation = PaymentSession.Operation(rawValue: operationType) else {
        throw InternalError(description: "Operation type is not known or supported: %@", operationType)
    }

    let isDeletable = isDeletable(allowDelete: listResult?.allowDelete, forOperationType: operationType)
    let context = PaymentContext(operationType: operation, extraElements: listResult?.extraElements, isDeletable: isDeletable)

    return .init(networks: translations.networks, accounts: translations.accounts, context: context)
}

/// Returns the defined deletable state according to https://optile.atlassian.net/browse/PCX-2012
private func isDeletable(allowDelete: Bool?, forOperationType operationType: String) -> Bool {
    switch operationType {
    // Return true if allowDelete is true or nil
    case "UPDATE":
        if (allowDelete == true) || (allowDelete == nil) {
            return true
        } else {
            return false
        }
    // Return true allowDelete is true; return false if allowDelete is nil or false
    case "CHARGE":
        // listResult.allowDelete is Optional
       return allowDelete == true
    default:
        return false
    }
}
```

PR #135, opened 2021-09-10
[PCX-2012] Respect allowDelete on UPDATE flow

caiomello:
**AS A** merchant
**I WANT** to offer my customers to add/edit their saved payment methods without allowing them to delete existing saved ones
**SO THAT** I always keep a list of my customers' saved payment methods

### Acceptance criteria
1. The delete payment method (trashcan) is available if and only if `allowDelete` in the LIST object is either set to `true` or undefined in UPDATE flow
2. The delete payment method (trashcan) is available if and only if `allowDelete` in the LIST object is set to `true` in CHARGE flow
3. The delete payment method (Trashcan) is only shown for Saved Accounts (AccountRegistration) and not for ApplicableNetworks or PresetAccounts.
---
Original PR: https://github.com/optile/checkout-ios/pull/132.


siebeprojects (dismissed):

Sources/UI/Model/PaymentContext.swift
```diff
@@ -10,9 +10,11 @@ import Foundation
 class PaymentContext {
     let listOperationType: PaymentSession.Operation
     let extraElements: ExtraElements?
+    let allowDelete: Bool?
```
Sorix, line 13: I remember that I proposed that change :).
I have the new proposal to avoid usage of `allowDelete` property in `PaymentContext` because we don't need it anywhere except initialization of `RegisteredAccount`, I don't think that we really should keep it.

You could pass directly `allowDelete` to `PaymentSession` without using context. That won't look super-nice but we will avoid filling context with data we won't use (at least with 3-states Boolean). And if we decide to use that Boolean in the future we anyway will convert it to the unwrapped Boolean.

_💡 It's the idea, not a blocker._

Sorix (approved):

Sources/UI/Model/PaymentContext.swift
caiomello: Addressed in https://github.com/optile/checkout-ios/pull/135/commits/58c0220c1a896cf011e22da57da07fce37d232f2. @siebeprojects please take a look.

caiomello: Hold on, not finished yet with this!

Sorix:
@caiomello unit tests are failing:

```
Testing failed:
	Tests:
		Extra argument 'allowDelete' in call
		'nil' requires a contextual type
		Missing argument for parameter 'allowDelete' in call
		Extra argument 'allowDelete' in call
		'nil' requires a contextual type
		Type of expression is ambiguous without more context
	Testing cancelled because the build failed.

** TEST FAILED **
```

caiomello:
> @caiomello unit tests [are failing](https://app.circleci.com/pipelines/github/optile/checkout-ios/869/workflows/b3c4a3ee-6590-43de-a542-94db8cb95bca/jobs/1954):
> 
> ```
> Testing failed:
> 	Tests:
> 		Extra argument 'allowDelete' in call
> 		'nil' requires a contextual type
> 		Missing argument for parameter 'allowDelete' in call
> 		Extra argument 'allowDelete' in call
> 		'nil' requires a contextual type
> 		Type of expression is ambiguous without more context
> 	Testing cancelled because the build failed.
> 
> ** TEST FAILED **
> ```

🤦🏻‍♂️ Fixed now. Tests are ✅.

siebeprojects (approved):

Sorix (approved):

PR #137, opened 2021-09-20
[PCX-2071] Write iOS functional tests for top and bottom extra elements

caiomello:
**AS** a mobile developer

**I WOULD** like that the extra elements functionality is covered by functional tests

**SO THAT** I know that any changes to the code base will not affect the extra element functionality.

**Acceptance Criteria**

The following tests should be written:

1. Test list result with only bottom elements. Click on clickable link and validate that URL is opened in external browser window.
2. Test list result with only top elements. Click on clickable link and validate that URL is opened in external browser browser window.
3. Test list result with both top and bottom elements. Click on top clickable link and validate URL is opened in external browser window.
4. Test list result with both top and bottom elements. Click on bottom clickable link and validate URL is opened in external browser window.
5. Test that ExtraElements with Checkbox are not shown in the payment page.


Sorix:
Blocked by backend implementation and a bug with registration options.

Sorix (commented):
When backend implementation will be done that branch would be modified to use network requests to the backend instead of GitHub static files, which will require some changes and I will review after that changes will be made.

ExampleCheckout/UITests/ExtraElementsTests.swift
```diff
@@ -0,0 +1,69 @@
...
+//
+//  Created by Caio Araujo on 16.09.21.
+//  Copyright © 2021 Payoneer Germany GmbH. All rights reserved.
+//
+
+import XCTest
+
+private let topBottomElementsURL = URL(string: "https://raw.githubusercontent.com/optile/checkout-android/PCX-2004/shared-test/lists/listresult_extraelements_topbottom.json")
+private let topElementURL = URL(string: "https://raw.githubusercontent.com/optile/checkout-android/PCX-2004/shared-test/lists/listresult_extraelements_top.json")
+private let bottomElementURL = URL(string: "https://raw.githubusercontent.com/optile/checkout-android/PCX-2004/shared-test/lists/listresult_extraelements_bottom.json")
```
Sorix, line 13: Before merging that should be deleted (we have a blocker for that, just as a reminder for myself).

ExampleCheckout/UITests/ExtraElementsTests.swift
```diff
@@ -0,0 +1,69 @@
...
+            XCTAssertFalse(app.textViews["Bottom Element Number 2 without Checkbox"].exists)
+        }
+    }
+
+    func testExtraElements_whenBottom_shouldDisplayBottom() throws {
+        try XCTContext.runActivity(named: "Open payment form") { _ in
+            try startPaymentSession(transaction: Transaction(), url: bottomElementURL)
+            app.tables.staticTexts["Cards"].tap()
+            XCTAssertTrue(app.textViews["Bottom Element Number 2 without Checkbox"].exists)
+            XCTAssertFalse(app.textViews["Top Element Number 1 with invalid link"].exists)
```
Sorix, line 48: Just for the sake of readability (top at the top, bottom at the bottom).

```suggestion
            XCTAssertFalse(app.textViews["Top Element Number 1 with invalid link"].exists)
            XCTAssertTrue(app.textViews["Bottom Element Number 2 without Checkbox"].exists)
```

_Not a blocker._

ExampleCheckout/UITests/ExtraElementsTests.swift
```diff
@@ -0,0 +1,69 @@
...
+            XCTAssertFalse(app.textViews["Bottom Element Number 1 with an Checkbox optional title"].exists)
+        }
+    }
+
+    func testExtraElements_whenContainsLink_shouldOpenSafari() throws {
+        try XCTContext.runActivity(named: "Open payment form") { _ in
+            try startPaymentSession(transaction: Transaction(), url: topBottomElementsURL)
+            app.staticTexts["Cards"].tap()
+            app.textViews.firstMatch.links["Number 1"].tap()
+            XCTAssertTrue(app.webViews.firstMatch.waitForExistence(timeout: 1))
```
Sorix, line 66: The thing I want to suggest is to increase the timeout because CI environment could be very laggy even in view controller loading time.

```suggestion
            XCTAssertTrue(app.webViews.firstMatch.waitForExistence(timeout: 5))
```

Sorix: @siebeprojects taught me not to use any non-named numbers in the code (sometimes I still fail with that). I don't know if it is applicable here but it could make the number more descriptive.

```suggestion
            let safariViewControllerPresentationTimeout = 5
            XCTAssertTrue(app.webViews.firstMatch.waitForExistence(timeout: safariViewControllerPresentationTimeout))
```

_Not a blocker (from my side)._

ExampleCheckout/UITests/PaymentSession/Model/Callback.swift
```diff
@@ -8,13 +8,13 @@ import Foundation
...
 
     /// URL of landing page in merchants shop system after customer select payment method. This property is mandatory for a `LIST` session with `operationType` of `PRESET`, or with deprecated `presetFirst` option set to `true`.
-    var summaryUrl: String?
+    let summaryUrl: String?
 
     /// Payment status notification URL. If defined, the OPG system will send asynchronous status notifications about this payment to this URL.
     ///
     /// - Note: merchant can configure a single notification URL for all transactions on the _division_ level via Merchant Configuration API. Notification URL in `callback`, however, overrides the division settings.
-    var notificationUrl: String?
+    let notificationUrl: String?
```
Sorix, line 19: That's a good topic to discuss. In structures, I often use `var` instead of `let`, because if you declare structure with `let` you couldn't change any values inside even if they are marked as `var`. But if you want to do it, you may declare it with `var` and change values inside.

I try to declare `let` when it could lead to bugs or where I'm sure that value shouldn't be changed. Here at some moment in the future, we could want to change it.

_Not a blocker._

ExampleCheckout/UITests/PaymentSession/Model/Customer.swift
```diff
@@ -8,10 +8,10 @@ import Foundation
 
 struct Customer: Codable {
     /// Customer identifier given by the merchant. Not validated for uniqueness by OPG.
-    var number: String
+    let number: String
 
     /// Customer e-mail address. It is highly recommended to provide it, since it is a mandatory information for some PSP and often used by advanced risk management.
-    var email: String?
+    let email: String
```
Sorix, line 14: Should be marked as optional to keep consistency with [a server model](https://www.optile.io/reference#operation/createPaymentSession).

```suggestion
    let email: String?
```

ExampleCheckout/UITests/PaymentSession/Model/Style.swift
```diff
@@ -11,5 +11,5 @@ struct Style: Codable {
     /// Preferred language for payment page. If undefined will be decided upon country information from transaction object.
     ///
     /// Format `<language code>[_<COUNTRY CODE>]`, where `<language code>` is a mandatory part that comply with [ISO 639-1 (alpha-2)](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes), and `<COUNTRY CODE>` is an optional part that comply with [ISO 3166-1 (alpha-2)](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2).
-    var language: String?
+    let language: String
```
Sorix, line 14: Should be marked as optional to keep consistency with [a server model](https://www.optile.io/reference#operation/createPaymentSession).

```suggestion
    let language: String?
```

Sorix (changes requested):
There are a few important comments; others are more about refactoring and styling. You could mark _non-blockers_ as resolved on your own without any comments if you think that your approach is good, it's absolutely ok.

ExampleCheckout/UITests/ExtraElementsTests.swift
caiomello: Of course! Probably a copy & paste mistake.

ExampleCheckout/UITests/ExtraElementsTests.swift
caiomello: > @siebeprojects taught me not to use any non-named numbers in the code

100% agree, magic numbers are bad. In test code I don't usually worry too much about this sort of thing, but we could have a "global" timeout constant for tests.

5 is A LOT though. If we have 100 tests each waiting 5 seconds before failing, it can add a lot of extra time to each test run. Whenever I've had to use more than 1 second timeout in a test, it was usually because something was wrong somewhere in the implementation. I can extract this but I'd keep it to 1 second and if a test fails because of that we can investigate or adjust it.

ExampleCheckout/UITests/NetworksTests.swift
```diff
@@ -10,19 +10,12 @@ class NetworksTests: XCTestCase {
     private(set) var app: XCUIApplication!
 
     /// Load an app and load networks list from list url.
-    /// - Parameter transaction: if `nil`, default `Transaction` will be used
-    func setupWithPaymentSession(using transaction: Transaction? = nil) throws {
+    func startPaymentSession(transaction: Transaction, url: URL? = nil) throws {
```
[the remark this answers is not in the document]
caiomello: The issue for me is the `WithPaymentSession`. With what payment session? It gives the impression that payment session is a parameter, but there's no "payment session" object.

It feels to me like the term "payment session" is being used for multiple purposes, so it gets pretty confusing.

Would `startPaymentProcess` make sense?

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
```diff
@@ -8,46 +8,33 @@ import Foundation
...
+    let transactionId: String
+    let country: String
+    let callback: Callback
+    let customer: Customer
+    let payment: Payment
+    let style: Style
+    let operationType: String
+    let allowDelete: Bool?
+
+    init(magicNumber: MagicNumber = .nonMagicNumber, operationType: OperationType = .charge, allowDelete: Bool? = nil) throws {
```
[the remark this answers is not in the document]
caiomello: That's what default parameters are for, it's equivalent to creating multiple initializers. The code that uses this function can just ignore the values they don't need to change. And those default/implicit values are declared right in the function declaration, so it's easy to see and change. I wouldn't worry about this argument list growing _now_. If it happens, then we solve it.

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
```diff
@@ -8,46 +8,33 @@ import Foundation
...
-
-    var division: String?
-
-    var allowDelete: Bool?
-}
-
-extension Transaction {
-    /// Load template transaction from JSON.
-    /// - Parameter amount: you could specify a custom amount (used as "magic number" for testing).
-    static func loadFromTemplate(amount: MagicNumber = .nonMagicNumber, operationType: OperationType = .charge) throws -> Transaction {
```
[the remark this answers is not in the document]
caiomello: Hmm what is that based on?

Initializers are just functions with some added behaviors, they exist to be the entry point for instantiation. There's no technical reason to use a static function here...

ExampleCheckout/UITests/PaymentSession/Model/Callback.swift
caiomello: There's a lot behind this topic... Let's see...

- Structs are immutable by default, and constants `let` are more performant and safer than mutable variables `var`
- Mutability and shared state are a huge source of problems in software and structs + constants are the biggest weapons Swift gives us against them
- Swift is very strongly in favor of immutability, and modern APIs have been moving in this direction more and more (see SwiftUI)

I could talk about this for a long time, but what I do is basically use structs and immutability by default. The code will perform better and be safer. In case I need some form of shared state or mutable properties, I use a class.

Happy to talk more about this stuff!

Sorix: If you define structure as immutable `let x = Callback()` all properties inside it will be immutable even they use `var` declaration.

caiomello: I think I misread your previous comment 😅. Yes it's technically the same if the instance is immutable, but using `var` opens it up for mutation. Why declare properties as mutable if you don't mutate them?

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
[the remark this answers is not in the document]
siebeprojects: Any operation that should not run in the UI/Main thread should not be done in initialiser/constructors either.
Additional behaviour/logic added to a constructor/initialiser makes it also hard to unit test. Initialisers/constructors should be kept simple because they could easily be used in the wrong place and that could lead e.g. to UI's to freeze.

It would be better to load the template file and create the Transaction Object in e.g. a function called fromTemplateFile() and this would return the Transaction Object. An additional comment could be added to the function that it should not run in the UI/Main thread.

ExampleCheckout/UITests/NetworksTests.swift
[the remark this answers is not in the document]
siebeprojects: Existing code and how things are named/implemented has already gone through a previous PR. So, before changing code/refactoring/renaming which is not part of the actual ticket, please discuss first before changing. It will save us time to not having to PR the same logic/code again.

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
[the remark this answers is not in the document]
siebeprojects: When initialisers/constructors have 5 - 6 or more parameters, it would be good to offload the construction to use e.g. a Builder pattern. When constructing/initialising an Object/Builder, it is a good practise to make sure it cannot be initialised with any invalid values (preconditions).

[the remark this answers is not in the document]
caiomello: I made changes to this on Friday and had to add 2 new parameters, and I agree it's too many. I'll work on this some more.

ExampleCheckout/UITests/NetworksTests.swift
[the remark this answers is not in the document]
caiomello: Sure, makes sense. This should be in the work agreement @Sorix.

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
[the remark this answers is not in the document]
caiomello: > Any operation that should not run in the UI/Main thread should not be done in initialiser/constructors either.
Additional behaviour/logic added to a constructor/initialiser makes it also hard to unit test. Initialisers/constructors should be kept simple because they could easily be used in the wrong place and that could lead e.g. to UI's to freeze.

This is running on the main thread, and I don't see any reason for it not to...

The context here is important: This is test code, so UI issues are irrelevant. In production code if I had to do something like this (generate data from a local file), I'd have an object responsible for file I/O with robust error handling and asynchronous operations. Doing something like that here would add complexity while not changing the functionality or bringing any practical benefits.

> Additional behaviour/logic added to a constructor/initialiser makes it also hard to unit test.

If this was doing any async work, yes. In this case I don't see how it would... this function only does two things:

1. Loads a template `Transaction` from a local file
2. Sets the initial value of each stored property

And it also throws (propagates errors up the chain), so if either the file fails to be loaded or the template `Transaction` can't be created, any code using this initializer will receive the relevant error and we'll always know what's going on.

> It would be better to load the template file and create the Transaction Object in e.g. a function called fromTemplateFile() and this would return the Transaction Object. An additional comment could be added to the function that it should not run in the UI/Main thread.

What would this accomplish? An initializer is just a single-purpose function. My goal here was to be consistent with how we use language features. Instantiation should be done via initializers. They provide extra safety and a standard syntax.

So far I've only received subjective opinions on this, but I'm really curious about the reasoning behind your suggestions. Did you have problems using constructors like this in the past? Was it a source of bugs? Did it negatively affect test code somehow?

ExampleCheckout/UITests/ExtraElementsTests.swift
Sorix: 5 seconds is not a lot for our CI. Sometimes just animating the presentation of the new view controller could take up to 10 seconds :), I caught timeout errors multiple times. You could extract it to some global property if you want, we just don't use it anywhere else, so I don't think it makes sense yet.

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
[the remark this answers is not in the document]
Sorix: > @sorix: That practice could lead to uncontrolled growth of an initializer in the future and we will need refactoring <...>

> @caiomello: I made changes to this on Friday and had to add 2 new parameters, and I agree it's too many. I'll work on this some more.

😇

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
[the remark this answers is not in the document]
caiomello: Addressed in https://github.com/optile/checkout-ios/pull/137/commits/ba99ffabd4f7cac16be9a19a5787fdd3d9995f3c.

ExampleCheckout/UITests/PaymentSession/Model/Transaction.swift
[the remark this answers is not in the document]
caiomello: Addressed in https://github.com/optile/checkout-ios/pull/137/commits/ba99ffabd4f7cac16be9a19a5787fdd3d9995f3c.

ExampleCheckout/UITests/NetworksTests.swift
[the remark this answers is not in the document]
caiomello: Addressed in https://github.com/optile/checkout-ios/pull/137/commits/b041fda928d50fd037a5e008e62eff76fd51d747.

ExampleCheckout/UITests/ExtraElementsTests.swift
caiomello: Addressed in https://github.com/optile/checkout-ios/pull/137/commits/ab1b24d09cbaf1048ba08ffba4ee2f2be2bb2187.

ExampleCheckout/UITests/PaymentSession/Model/TransactionSettings.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  TransactionSettings.swift
+//  UITests
+//
+//  Created by Caio Araujo on 04.10.21.
```
siebeprojects, line 5: Please do not use author names or dates in source files, use the default header we use similar in other file headers.

caiomello: 🤦🏻‍♂️

caiomello: I also added to the example project the same template file we have in the main project (https://github.com/optile/checkout-ios/pull/137/commits/dbf5a7e44e2cc149b16fc5795b9b9c5691978a9d).

Sorix (approved):

ExampleCheckout/UITests/PaymentSession/Model/TransactionSettings.swift
```diff
@@ -0,0 +1,29 @@
...
+        division: String? = nil,
+        checkoutConfiguration: CheckoutConfiguration? = nil,
+        allowDelete: Bool? = nil
+    ) {
+        self.magicNumber = magicNumber
+        self.operationType = operationType
+        self.division = division
+        self.checkoutConfiguration = checkoutConfiguration
+        self.allowDelete = allowDelete
+    }
```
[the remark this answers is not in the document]
caiomello: Did you mean

```swift
var magicNumber: Transaction.MagicNumber = .nonMagicNumber
var operationType: Transaction.OperationType = .charge
var division: String? = nil
var checkoutConfiguration: CheckoutConfiguration? = nil
var allowDelete: Bool? = nil
```

?

You're right that it would be work the same, yes, but then you allow mutability, which in general means 3 things:

1. **Inconsistency** by allowing properties to be set in two different ways, via `init` and after it:

```swift
let transaction = Transaction(allowDelete: true)

var transaction = Transaction()
transaction.allowDelete = true
```

2. **Worse performance**: by declaring a property as immutable you tell the compiler that this value is never going to be changed and it can optimize for that

3. **Decreased safety**: allowing values to be changed after initialization anywhere in the code makes it harder to keep track of state, which in turn makes the code harder to think about and debug

This has been an established good practice since the first version of Swift:

image

https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html

Even if this is test code (and I don't think we need to be as strict with it as we do with "product" code), I think immutability is a really important characteristic of Swift that we should take advantage of.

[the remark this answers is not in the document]
caiomello: I'm breaking protocol and resolving this one myself since @Sorix is out today. This PR has been open for too long already...

PR #146, opened 2021-10-04
[PCX-1653] Refine caching

Sorix:


Sources/UI/Scenes/List/PaymentListViewController.swift
```diff
@@ -69,6 +69,12 @@ extension PaymentListViewController {
     public override func didRotate(from fromInterfaceOrientation: UIInterfaceOrientation) {
         methodsTableView?.reloadData()
     }
+    
+    public override func viewWillDisappear(_ animated: Bool) {
+        // Follow Android SDK caching rules, clear all caches on exit
+        sessionService.connection.cache?.removeAllCachedResponses()
```
Sorix, line 75: That should be changed in the future (for example use it in `deinit` of `sessionService`). I can't do that right now because I found that we have memory issues and `deinit` is not being called in the list view controller (so won't be called in session service), which should be fixed in another ticket and we could return to this.

PR #153, opened 2021-11-02
[PCX-2308] GitHub labels

Sorix:
**AS A** developer
 **I WANT** to have better readability of pull requests
 **SO THAT** labels should be automatically assigned for pull requests

### Acceptance criteria
Labels should be assigned automatically via GitHub actions.
1. On Browserstack deployment `awaits qa` label should be assigned.
1. For bug tickets `bug` label should be assigned.
1. For feature tickets `feature` label should be assigned.


.github/workflows/publish_app_live.yml
```diff
@@ -89,6 +93,13 @@ jobs:
...
-        with:
-          labels: awaits qa
\ No newline at end of file
+  apply-label:
+     runs-on: ubuntu-latest
+     needs: upload-to-browserstack
+
+     steps:
+       - uses: actions-ecosystem/action-add-labels@v1
+         if: ${{ github.event.inputs.pull-request-number != '' }}
```
caiomello, line 102: What does this condition mean? How are QA labels assigned?

caiomello (approved):

.github/workflows/publish_app_live.yml
Sorix: That means if the pull request number is not defined action won't run. You could run that action only manually, because of that it couldn't detect PR's number for the branch which was used to run that action to assign the label.

Sorix: So when you run that action, you could specify PR's number to assign the label.

caiomello: I just noticed the file name `publish_app_live.yml`, is that from GitHub? We don't publish any apps...

Sorix: We publish the app to Browserstack, @siebeprojects proposed to name it `publish`, I used `deploy` previously.

bzlat (approved):

PR #154, opened 2021-11-08
[PCX-2119] Replace Material components

caiomello:
> [PCX-2119](https://optile.atlassian.net/browse/PCX-2119)

- Integrated new text field component with existing logic and replaced Material text fields in
  - `Input.Table.TextFieldController`
  - `Input.Table.TextFieldViewCell`
  - `Input.Table.CVVTextFieldViewCell`
- Removed the Material components library

| Idle | Error |
| - | - |
|  |  |


Sources/UI/Components/TextInputView/CALayer+Extensions.swift
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2021 Payoneer Germany GmbH
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import UIKit
+
+extension CALayer {
```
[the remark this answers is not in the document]
caiomello: [Core Animation](https://developer.apple.com/documentation/quartzcore), the underlying animation framework on Apple OSs.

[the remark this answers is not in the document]
Sorix: Old-style modules in iOS use prefixes, like *NS*String, *UI*TableView, *CA*Layer, etc.

Sources/UI/Scenes/Input/Table/Cell/Input.Table.TextFieldViewCell.swift
```diff
@@ -15,27 +14,27 @@ extension Input.Table {
...
             textFieldController.delegate = self
 
             // Add the text field to a view
-            contentView.addSubview(textField)
-            textField.translatesAutoresizingMaskIntoConstraints = false
+            contentView.addSubview(textInputView)
+            textInputView.translatesAutoresizingMaskIntoConstraints = false
 
-            let textFieldBottomAnchor = textField.bottomAnchor.constraint(lessThanOrEqualTo: contentView.bottomAnchor)
+            let textFieldBottomAnchor = textInputView.bottomAnchor.constraint(lessThanOrEqualTo: contentView.bottomAnchor, constant: -10)
```
Sorix, line 31: Could you name `-10`?

Sorix (changes requested):
1. I could say that animation issues are worse than I expected looking at videos, I don't know if we could merge it how it looks now (at least on iOS 12.4). We should discuss that if it is impossible to resolve them in `UICollectionView` (is it?) we should move to `UIStackView` before merging that to develop branch.

    https://user-images.githubusercontent.com/5610904/142144088-aec9fa06-3b48-4aa3-af3d-13339f220689.mov


1. Hint button doesn’t match the design’s color (I think something wrong with the alpha component)
1. Clear button color doesn’t match hint button’s color.
1. Margins or paddings for the clear and the hint buttons are not the same.

Sorix (approved):

Sources/UI/Components/TextInputView/TextInputView.swift
```diff
@@ -139,11 +142,12 @@ extension TextInputView {
...
-        if animated {
-            animate(changes)
-        } else {
+        // Animations are turned off for now so they don't conflict with UICollectionView reload animations.
+//        if animated {
+//            animate(changes)
+//        } else {
             changes()
-        }
+//        }
```
Sorix: Last time @siebeprojects reviewed my code, he asked not to commit commented code blocks. If you want still to commit with comments, there is no need in constant `animated` above https://github.com/optile/checkout-ios/blob/e29e328c99c54fa0c3839a0ec7882d857a39f6ef/Sources/UI/Components/TextInputView/TextInputView.swift#L105

Sorix: Isn't it better to remove `animated` argument from `setStatus` method. And add `- TODO: ` note to set it later when we migrate to `UIStackView`?
https://github.com/optile/checkout-ios/blob/e29e328c99c54fa0c3839a0ec7882d857a39f6ef/Sources/UI/Components/TextInputView/TextInputView.swift#L97-L101

Sorix: By the way, when I checked that change on a simulator I didn't notice any visual changes.

Sources/UI/Components/TextInputView/TextInputView.swift
```diff
@@ -101,6 +101,9 @@ extension TextInputView {
     }
 
     private func updateAppearance(animated: Bool) {
```
Sorix, line 103: Isn't it better to remove `animated` argument. And add `- TODO: ` note to set it later when we migrate to `UIStackView`? If I correctly understand the change.

```suggestion
    // - TODO: return `animated` argument when we resolve issues with UICollectionView animation (see: https://github.com/optile/checkout-ios/pull/154#pullrequestreview-807716808)
    private func updateAppearance() {
```

caiomello: I see two options here:

1. We keep the current code as intact as possible while disabling animations (the way it is now)
2. We remove all animation code as if it never existed in the first place

I chose 1 because I think this code is really valuable, I really want to have this enabled again and the UI working really well as it should, and also I don't know when we'll be able to get it done and it'd be great if when the time comes we can just change these 2 small things and get the original functionality back.

@siebeprojects @Sorix I'm looking forward to hearing your opinions on this.

siebeprojects: @caiomello when the TextInputView has the option to either enable/disable animations then we should use that functionality ourselves to disable animations if we do not want it at the moment.
Either we set the Default value to "false", or we initialise all TextInputViews with the animation flag to false.

Sources/UI/Components/TextInputView/TextInputView.swift
```diff
@@ -0,0 +1,231 @@
...
+        let shouldHideError = configuration.errorMessage.isEmpty
+
+        let changes = { [weak self] in
+            self?.alpha = self?.status == .disabled ? 0.5 : 1
+            self?.errorLabel.isHiddenInStackView = shouldHideError
+            self?.errorLabel.alpha = shouldHideError ? 0 : 1
+            self?.errorLabel.superview?.layoutIfNeeded()
+        }
+
+        // Animations are turned off for now so they don't conflict with UICollectionView reload animations.
```
siebeprojects, line 145: There is no need to disable / comment this code because we could just initialise every TextInputView with the animation flag set to false.

Sources/UI/Components/TextInputView/TextInputView.swift
caiomello: You are absolutely correct @siebeprojects! That's definitely the way to go. The component provides the functionality, it's up to the example app to decide if it animates or not of course. I don't know how that didn't occur to me 🤦🏻‍♂️

Sources/UI/Components/TextInputView/TextInputView.swift
```diff
@@ -0,0 +1,230 @@
...
+// MARK: - UITextFieldDelegate
+
+extension TextInputView: UITextFieldDelegate {
+    func textFieldDidBeginEditing(_ textField: UITextField) {
+        updateAppearance(animated: shouldAnimate && textField.frame != .zero) // Prevent animation while view is being laid out
+        delegate?.textInputViewDidBeginEditing(self)
+    }
+
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        setStatus(.normal)
```
Sorix, line 218: What's the purpose of that call?

Sorix (commented):
#### QA
- [ ] Return keys on a keyboard (_Next_, _Return_) seems don't work

Sources/UI/Components/TextInputView/TextInputView.swift
caiomello: Removing the error state when text changes. We might not need it though because state is being set by the delegate when focus changes instead... I'll double check.

Sorix: Yes, we remove that status when text field receives focus outside of the component, so there is no need to reset an error state because in that case we have duplication.

Sorix (approved):

siebeprojects (approved):

PR #155, opened 2021-11-11
[PCX-995] Display PRESET method in payment list

Sorix:
**AS AN** end-customer
**I WANT** to see my previously selected payment method within the payment list (alongside my saved accounts and other available payment methods)
**SO THAT** I can use IT without having to fill its details again

### Acceptance criteria
1. When a LIST object contains a `preset` account, a new section appears at the top of the payment list with the title “Previously selected” (`networks.preset.title`)
1. If the preset account is not a registered one, neither is intended to be registered in the current session, a warning text (`networks.preset.conditional.text`) appears as a subtitle beneath the title. This happens if all three flags (`presetAccount.isRegistered`, `presetAccount.autoRegistration`, `presetAccount.allowRecurrence`) in the `presetAccount` object are set to false.
1. The display of the account is similar to that of saved account and is specified in the linked design ticket
1. The operation button for this previously selected method has the `button.operation.PRESET.label` text, which matches the `operationType` parameter in the LIST object for the `preset` network in the LIST object

### Remarks
* Clicking the “Continue” button is not in the scope of this ticket.


ExampleCheckout/UITests/PaymentSession/Model/Server/PresetAccount.swift
```diff
@@ -20,7 +20,7 @@ class PresetAccount: NSObject, Decodable {
     let emptyForm: Bool
 
     /// Code of button-label for this preset account.
-    let button: String
+//    let button: String?
```
siebeprojects, line 23: Is this code still needed or not?

Sorix: I think yes because [documentation](https://www.optile.io/reference#operation/getPaymentSession) still says the button is a required field and should be present in a model, but a server doesn't return it so I receive a JSON parsing error.

I think I should remove (or revert a comment) it when documentation will be updated or the server will be fixed. It's my idea, but I could easily remove that string.

siebeprojects: I prefer not having commented code because at the end we dont know if it is still needed or not.

Sources/UI/Scenes/List/Table/Cells/List.Table.DetailedHeaderView.swift
```diff
@@ -0,0 +1,79 @@
...
+                    view.trailingAnchor.constraint(equalTo: contentView.layoutMarginsGuide.trailingAnchor)
+                ]
+            }
+
+            NSLayoutConstraint.activate(constraints)
+        }
+    }
+}
+
+private extension UIFont {
```
caiomello, line 67: Again these private extensions really worry me... We should move away from this practice, but it's out of the scope of this PR. I'll open a topic in the maintenance board so we can discuss.

caiomello (dismissed):
I don't see any big issues, so my comments are just pointing out some small things.

Doing these big "clean up" changes (`UIModel` + `Dequeuable`) that are kind of out of the scope of the task makes it much harder to review PRs. These sorts of changes should be separate tasks/PRs, but because we don't currently have time allocated for maintenance work I understand it's basically the only way to get this done. This is a good topic to bring up maybe in the next retro.

Sources/UI/Model/UIModel.PaymentSession.swift
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2021 Payoneer Germany GmbH
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Foundation
+
+extension UIModel {
```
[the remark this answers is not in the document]
Sorix: We should, but it's a part of another long story :)

Sources/UI/Protocols/Dequeueable.swift
```diff
@@ -8,47 +8,56 @@
 
 import UIKit
 
-protocol DequeueableCell: AnyObject {
+protocol Dequeueable: AnyObject {
```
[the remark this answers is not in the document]
Sorix: I don't understand why it is not necessary, it protects from registering or dequeuing incorrect classes.

Sources/UI/Scenes/List/Table/Cells/List.Table.DetailedHeaderView.swift
Sorix: You could open a discussion, I don't see any problems with private extensions.

Sources/UI/Protocols/Dequeueable.swift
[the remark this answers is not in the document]
caiomello: I'm not sure what you mean by "unreliable". [This](https://github.com/caiomello/utilities/blob/main/Sources/Utilities/Extensions/UITableView%2BExtensions.swift) is what I'm talking about.

Sources/UI/Model/UIModel.PaymentSession.swift
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            // Preset account
+            if let translatedPresetAccount = presetAccount {
+                let warningText: String?
+                if Self.shouldDisplayWarningText(for: translatedPresetAccount.model) {
+                    warningText = translatedPresetAccount.translator.translation(forKey: "networks.preset.conditional.text")
+                } else {
+                    warningText = nil
+                }
```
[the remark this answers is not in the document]
Sorix: I'm not sure that it is more readable O_O

Sorix:
> @caiomello: Doing these big "clean up" changes (UIModel + Dequeuable) that are kind of out of the scope of the task makes it much harder to review PRs. These sorts of changes should be separate tasks/PRs, but because we don't currently have time allocated for maintenance work I understand it's basically the only way to get this done. This is a good topic to bring up maybe in the next retro.

I had to do it because of `PresetAccount`, otherwise, it conflicts (server model and UI model).
I also had to modify `Dequeuable` because I use it in a `UITableHeaderFooterView` (previously I used it only in table and collection views).

ExampleCheckout/UITests/PaymentSession/Model/Server/PresetAccount.swift
Sorix: I changed in 74b0626daef677a0770164c521c3b3ca4ed5d81c. Warning could be removed after merging https://github.com/optile/docs-rest-apis/pull/250.

ExampleCheckout/Sources/Swift/Theme+custom.swift
```diff
@@ -13,6 +13,7 @@ extension Theme {
             backgroundColor: .white,
             tableBorder: .init(red: 0.56, green: 0.58, blue: 0.60, alpha: 1),
             tableCellSeparator: .init(red: 0.94, green: 0.95, blue: 0.95, alpha: 1),
+            tablePresetBordersColor: .blue,
```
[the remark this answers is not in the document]
Sorix: Changed to a tint color: f7e19bcd03fae44d1a6a10a00a379026e1aa0d4e

Sources/UI/Protocols/Dequeueable.swift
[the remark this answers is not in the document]
Sorix: I meant that using textual representation like `\(Cell.self)` is not reliable, that behavior could be changed in the future and doesn't guarantee an alphanumeric textual identifier. I try to use specialized and documented methods to avoid such situations, that's why I use a helper `NSStringFromClass`.
> Returns the name of a class as a string.

Sources/UI/Model/UIModel.PresetAccount.swift
```diff
@@ -0,0 +1,53 @@
...
+
+            self.networkLabel = localizer.translation(forKey: "network.label")
+            self.submitButtonLabel = localizer.translation(forKey: submitButtonLocalizationKey)
+
+            logo = Loadable<UIImage>(identifier: apiModel.code.lowercased(), url: apiModel.links["logo"])
+        }
+    }
+}
+
+extension UIModel.PresetAccount {
```
siebeprojects, line 32: This looks like a duplication from the code in the UIModel of the RegisteredAccount. Why not use a shared function that generates a readable label from a MaskedAccount given the name of the input element and probably later the type of Payment Method like CREDIT_CARD, WALLET, DEBIT_CARD.

Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.RegistrationOptionsBuilder.swift
```diff
@@ -22,8 +22,7 @@ extension Input.ModelTransformer {
 extension Input.ModelTransformer.RegistrationOptionsBuilder {
     func createInternalModel(fromRegistration registration: ApplicableNetwork.RegistrationOption, reccurrence: ApplicableNetwork.RegistrationOption) throws -> [InputField] {
         switch listOperationType {
-        // That case applies to PRESET flow as well when it will be supported
-        case .CHARGE:
+        case .CHARGE, .PRESET:
             return try createInputFields(forChargeFlowUsingRegistration: registration, recurrence: reccurrence)
```
siebeprojects, line 26: Would  "forDefaultFlowUsingRegistration" be a better name now that it is used for both CHARGE and PRESET flow.

Sources/UI/Model/UIModel.PaymentSession.swift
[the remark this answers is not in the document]
caiomello: The line breaks on GitHub make it look pretty bad... I see a lot of benefit in isolating value creation within a single block, but as you wish. In this case it's not that bad because it's in an initializer and you don't need to juggle mutable values. I agree with you about `guard` to a certain extent, I've seen _a lot_ of guard overuse and it can be terrible.

Sources/UI/Service/PaymentSessionService/PaymentSessionProvider.swift
```diff
@@ -140,28 +138,40 @@ class PaymentSessionProvider {
...
 
-        let context = PaymentContext(operationType: operation, extraElements: listResult?.extraElements)
+        let context = UIModel.PaymentContext(operationType: operation, extraElements: listResult?.extraElements)
 
-        return .init(networks: translations.networks, accounts: translations.accounts, context: context, allowDelete: listResult?.allowDelete)
+        return .init(networks: translations.networks, accounts: translations.accounts, presetAccount: translations.presetAccount, context: context, allowDelete: listResult?.allowDelete)
     }
 }
+
+private struct APINetworks {
```
siebeprojects, line 173: Would ListNetworks, SupportedNetworks or FilteredNetworks be a better name than APINetworks

Sources/UI/Model/UIModel.PresetAccount.swift
Sorix: I wanted to do it later when we come to an agreement with a designer about how masked account labels should look in different scenarios (product owner asked to make a meeting about this topic, I will schedule it). After that agreement, I wanted to create a shared protocol that will work in different scenarios.

I could do it in that ticket, but I wanted to implement that separately when I have all the needed information.

Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.RegistrationOptionsBuilder.swift
Sorix: I'm not sure, did you highlight the right line? Do you talk about enumeration case iterator or do you talk about `createInternalModel` named arguments?

Sources/UI/Protocols/Dequeueable.swift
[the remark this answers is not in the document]
caiomello: > I meant that using textual representation like `\(Cell.self)` is not reliable, that behavior could be changed in the future and doesn't guarantee an alphanumeric textual identifier

1. We build on top of a platform, _everything_ can be changed in the future, it's our job to keep things working as the underlying code changes. How is this any different?
2. The approach I linked to is something I've been using for many years, in all sorts of different projects, big and small. It has always been 100% reliable so I'm really curious to know if I'm missing something

Using the `ViewController` class in the `TextFieldDemo` project as an example:

```swift
"\(ViewController.self)"
String(describing: ViewController.self)
NSStringFromClass(ViewController.self)
```

They return, respectively:

```
ViewController
ViewController
TextFieldDemo.ViewController
```

**The first two options are the modern ones, and we should use them.**

This is your helper:

```swift
private extension NSObject {
    class var nameOfClass: String {
        let fullClassName = NSStringFromClass(self)
        guard let className = fullClassName.components(separatedBy: ".").last else {
            assertionFailure("Using unexpected class name (full)")
            return fullClassName
        }

        return className
    }
...
```

If I was reviewing a PR containing this code I'd not accept it.

1. There's no reason to use `NSStringFromClass` over the other options if you're discarding the module name
2. You're doing string manipulation to get a simple class name
3. This is behind yet another private extension. If someone else needs something like this somewhere else they'll either write the same code in another place, write a different implementation to do the same functionality, or use another method altogether, adding inconsistency and making the codebase worse overall
4. If this fails to return the expected class name for a UI element it will result in an undefined UI state in a release build, which is, in most cases, _worse_ than stopping execution

caiomello:
> I had to do introduce UIModel because of PresetAccount, otherwise, it conflicts (server model and UI model).

If the addition of a new type requires such a large change, then there's a deeper problem here. We should try to understand the root cause (maybe you already do, I'm happy to learn more) and solve it so this doesn't keep happening.

> I also had to modify Dequeuable because I use it in a UITableHeaderFooterView (previously I used it only in table and collection views).

I got that and I agree with the name change 👍

caiomello (approved):

Sources/UI/Protocols/Dequeueable.swift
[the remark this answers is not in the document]
Sorix:
```swift
\(ViewController.self)
String(describing: ViewController.self)
```

It’s almost the same call. when you ask for the textual representation, `\()` will use `String(describing:)`.

> The first two options are the modern ones, and we should use them.
The behavior for that options is not documented unlike the old function, so anyway “it works in the past” is not an argument. They just return that name, but it could return anything else.

I recommend reading that answer for example https://stackoverflow.com/a/55208499/4532985, I agree that usage of a specific tool (and documented one!) is better than using a generic one created with other purposes.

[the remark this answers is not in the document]
Sorix: One more link: https://forums.swift.org/t/relying-on-string-describing-to-get-the-name-of-a-type/16391/2 (I don't know if it is Apple's developer).

Sources/UI/Service/PaymentSessionService/PaymentSessionProvider.swift
Sorix: Resolved in c847f6abd6bb84568e9409de3d52999f8068eff6

Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.RegistrationOptionsBuilder.swift
siebeprojects: `createInputFields(forChargeFlowUsingRegistration:` the `forChargeFlowUserRegistration` could be called forDefaultFlowUserRegistration because it applies both to Preset and Charge flows.

Sources/UI/Model/UIModel.PresetAccount.swift
siebeprojects: Ok, we could do that in another ticket.

Sources/UI/Scenes/Input/Model/ModelTransformer/Input.ModelTransformer.RegistrationOptionsBuilder.swift
Sorix: Changed in 4357d13d25696bc0b23c12b35e6ceb1e8ecd6574

caiomello (approved):

Sources/UI/Protocols/Dequeueable.swift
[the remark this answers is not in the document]
caiomello: Very interesting! Thanks for sharing.

> “it works in the past” is not an argument

Our past experiences inform the way we do things today. The fact that I've done something for years at the very least gives some degree of confidence about how reliable it is. It might not be 100% reliable forever (nothing is), but it is _definitely_ reliable enough to be used in real code.

As always, it's important to take context into consideration. The argument about the string method being unreliable is pretty much irrelevant in our case, we're not building something that would require that level of "perfection", and like I said we build on top of a platform, and that platform changes all the time. Breaking changes and deprecations are part of software development. There's nothing special about `NSStringFromClass`. If we extrapolate that logic than we would not build on top of iOS and instead make a whole OS from scratch so we don't run the risk of things changing outside of our control.

Those links raise very good points, but `NSStringFromClass` is part of the Objective-C runtime and is limited to class types only. I personally wouldn't use it over simple string interpolation because of the limitations and the additional overhead of having to parse the name from the final string.

I'm completely OK to keep the code like this. I think the worst aspect of it by far is that it's a private extension, but that's already an ongoing topic 👍 

All this being said, if we go back to my first comment, it still stands that no matter what method you use to get the class name, the `Dequeueable` protocol is redundant because you can simply replace `Cell.identifier` with `Cell.nameOfClass` in the extension methods.

siebeprojects (approved):

PR #156, opened 2021-11-22
[PCX-992] Pre-set payment method or saved account

Sorix:
AS AN end customer
I WANT to preset one of my saved accounts or a newly entered account
SO THAT I can later use it to pay and finalise my order after I’ve seen its summary

### Acceptance criteria
1. The CTA (call to action) button in the case of `PRESET` flow (can be recognised from the LIST object’s `operationType` parameter) displays “Continue” (for the English language) - `button.operation.PRESET.label`
1. Clicking the button triggers the `operation` URL of the selected network or saved account, ending in `…/preset`.
  1. Payment result is returned to merchant scope. In the scope of the example app, the interaction code/reason and readable text appear just as they would for a CHARGE operation

### Changelog
- [x] Rename `ChargeRequest` to `OperationRequest` because it is not only responsible for _charge_ operations.
- [x] Move all network request to a separate namespace to avoid name collisions with UI's `OperationRequest`

#### Notes
* This ticket’s scope does not include an already pre-set account - only saved accounts and available networks. Handling an already pre-set account is done in the scope of [PCX-996: Pre-set an already PRESET account](https://optile.atlassian.net/browse/PCX-996)


Sources/Network/Requests/.swiftlint.yml
```diff
@@ -0,0 +1,2 @@
+disabled_rules:
+  - nesting
```
caiomello, line 2: This is because of the namespace/extension thing?

caiomello (approved):

Sources/Network/Requests/.swiftlint.yml
Sorix: That's because `typealias` nested 2 levels deep, SwiftLint by default allows only 1 level deep. I could define `typealias` in extension but that won't look nice, so I chose to disable SwiftLint rule in that case.

https://github.com/optile/checkout-ios/blob/60da8fc45cfe4b856cad99e1a9ee67dc8abed2dd/Sources/Network/Requests/DownloadData.swift#L20

caiomello (dismissed):

siebeprojects (approved):

caiomello (approved):

PR #157, opened 2021-11-23
[PCX-996] Pre-set an already PRESET account

Sorix:
**AS AN** end-customer
**I WANT** to select my previously selected method and proceed to the summary page
**SO THAT** I can complete my payment with this method

### Acceptance Criteria

1. Clicking the PresetAccount UI card brings up the a bottom sheet with a “Continue” button (`button.preset.label`). This could be improved in the future by introducing clickable cards.
1. When the “Continue” button is clicked, the ios-sdk returns a PaymentResult to the merchant including a client-side created OperationResult, Interaction and Redirect object (from PresetAccount). Note that `operation` (in this case a `/charge` is not called!
1. The PresetAccount object stored in the ListResult has a Redirect object. This Redirect object contains the interactionCode and interactionReason parameters and these are used to create the Interaction Object.
1. The Redirect object from the PresetAccount object is also stored in the newly created OperationResult object.
1. The OperationResult will not have any providerParameters.
1. If the Redirect object does not contain an interactionCode or interactionReason parameter than an error is returned to the user with Interaction (ABORT, CLIENTSIDE_ERROR), the resultInfo will be “Missing Interaction code and reason inside PresetAccount.redirect”.

The android-sdk implementation for selecting an already selected preset-account can be found here:
https://github.com/optile/checkout-android/blob/238c7565e4b04625fb89acc0dc093f4a63a5e770/checkout/src/main/java/com/payoneer/checkout/ui/page/PaymentListPresenter.java#L512-L527


Sources/UI/Scenes/Input/Input.ViewController.PaymentController.PresetResponseBuilder.swift
```diff
@@ -0,0 +1,58 @@
...
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Foundation
+
+extension Input.ViewController.PaymentController {
+    /// Manual response builder for pre-set an already PRESET account
+    /// - SeeAlso: https://optile.atlassian.net/browse/PCX-996
```
caiomello, line 11: Typo?

```suggestion
    /// - See also: https://optile.atlassian.net/browse/PCX-996
```

caiomello (approved):

Sources/UI/Scenes/Input/Input.ViewController.PaymentController.PresetResponseBuilder.swift
Sorix: It's XCode markup language [Markup Formatting Reference: See Also](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/SeeAlso.html).

siebeprojects (approved):

PR #158, opened 2021-11-25
[PCX-2363] Add support for Swift Package Manager

caiomello:
# Changes

- New file structure (screenshots below)
- How we access bundles (See `Bundle+Extensions.swift`)
- Renamed `Assets` folders to `Resources` to match SPM's standard structure
- Updated `fastfile`
- Updated GitHub Actions
- Updated `podspec`

# What to review

- Usage of `Bundle`
- New folder structure and files (screenshots below)
- `fastfile`
- `.github/pullrequest.yml`

# Demo

| Package | Example app |
| - | - |
|  |  |


.github/workflows/pullrequest.yml
```diff
@@ -1,12 +1,15 @@
 name: Unit and UI tests
 
 on:
+  push:
+    branches:
+      - feature/PCX-2363-spm
```
Fbada006, line 6: Why are you adding this here? I do not see why a feature branch needs to be added to the yml file. This means you will have to add something similar every time you have a new branch? You only need to do this for `develop` and/or `main` branches only unless I am missing something

caiomello: This was for testing. I just removed it, good catch!

Sources/Core/Bundle+Extensions.swift
```diff
@@ -0,0 +1,22 @@
...
+import Foundation
+
+extension Bundle {
+    /// Returns `module` when the SDK is being imported via SPM. Returns `main` when being used through other ways (e.g. CocoaPods).
+    static var current: Bundle {
+#if SWIFT_PACKAGE
+        return .module
+#else
+        return .main
+#endif
```
Sorix, line 16: Is it possible to indent `#if` statement, will it work?

Sources/UI/Model/Loadable.swift
```diff
@@ -20,8 +20,7 @@ enum Loadable<T> {
 extension Loadable where T == UIImage {
     /// Initialize loadable image object, it will be set to `.loaded` if local asset with such identifier exists. Returns `nil` if no local asset was found and URL wasn't specified.
     init?(identifier: String, url: URL?) {
-        let bundle = Bundle(for: UIModel.PaymentNetwork.self)
-        if let image = UIImage(named: identifier, in: bundle, compatibleWith: nil) {
+        if let image = UIImage(named: identifier, in: Bundle.current, compatibleWith: nil) {
```
Sorix, line 23:
```suggestion
        if let image = UIImage(named: identifier, in: .current, compatibleWith: nil) {
```

Gemfile.lock
```diff
@@ -333,4 +333,4 @@ DEPENDENCIES
   fastlane-plugin-browserstack
 
 BUNDLED WITH
-   2.2.3
+   2.2.22
```
Sorix, line 336: Seems like you're using lower Bundler version, could you update it to the latest one?

Package.swift
```diff
@@ -0,0 +1,36 @@
...
+        .iOS(.v12)
+    ],
+    products: [
+        .library(
+            name: "PayoneerCheckout",
+            targets: ["PayoneerCheckout"])
+    ],
+    dependencies: [
+
+    ],
```
Sorix, line 18: Is it needed?

PayoneerCheckout.podspec
```diff
@@ -8,14 +8,8 @@ Pod::Spec.new do |s|
...
   s.source           = { :git => 'https://github.com/optile/checkout-ios.git', :tag => s.version.to_s }
 
-  s.ios.deployment_target = '12.4'
-  s.swift_version = '5.3'
+  s.ios.deployment_target = '12.0'
+  s.swift_version = '5.5'
 
   s.source_files = 'Sources/**/*.swift'
-  
-  s.resources = ['Sources/Assets/*']
```
Sorix, line 16: I think we need to specify the new path to resources here, why did you delete it?

```suggestion
  s.resources = ['Sources/Resources/*']
```

PayoneerCheckout.podspec
```diff
@@ -8,14 +8,8 @@ Pod::Spec.new do |s|
...
-
-  # s.public_header_files = 'Pod/Classes/**/*.h'
-  s.frameworks = 'UIKit'
```
Sorix, line 19: Why did you delete this line?

```suggestion
  s.frameworks = 'UIKit'
```

caiomello: AFAIK this causes CocoaPods to include UIKit in the global header, meaning all files in the SDK implicitly import UIKit. Everything works fine after removing it, but there's very little documentation on this, so if you have more info let me know.

Package.swift
caiomello: I usually like to leave it in because it's part of the generated syntax and we'll most likely use it soon, but I can remove it no problem.

PayoneerCheckout.podspec
caiomello: Good catch!

Sources/Core/Bundle+Extensions.swift
caiomello: Manually yes, but it's common practice to have pre-processor macros on column 0, and Xcode indents like this automatically.

PayoneerCheckout.podspec
caiomello: I believe this is only needed if you want to expose resources to the host app. Everything works after removing it, but I'll  test some more.

Sorix: The framework automatically downloads images when they're not found locally. Please check, if the framework is not downloading it from the internet after disabling this, it is okay.

PayoneerCheckout.podspec
Sorix: At least here (https://guides.cocoapods.org/syntax/podspec.html#frameworks) I see that we should include all system framework we use. Despite the fact that I believe that everything can work without it I think it's safer to include it.

Package.swift
Sorix: Just doesn't look nice :), but that is a matter of taste :). I will resolve that comment because it is more about decoration.

```suggestion
dependencies: [],
```

Gemfile.lock
Sorix: Resolved in f17003a

Package.swift
caiomello: I removed it already, no worries.

PayoneerCheckout.podspec
Sorix: You could add a breakpoint here to check this:
https://github.com/optile/checkout-ios/blob/4568c6205bea169959be3f59fbb4011e281db7fe/Sources/UI/Service/PaymentSessionService/DataDownloadProvider.swift#L30-L37

And check output as well it should be triggered as well: https://github.com/optile/checkout-ios/blob/4568c6205bea169959be3f59fbb4011e281db7fe/Sources/Network/Requests/DownloadData.swift#L31

caiomello: Updated in https://github.com/optile/checkout-ios/pull/158/commits/7e4d918829d1992e9b66075d4c4b2202e58e82a3. Thanks for pointing this out and for the thorough review!

Sorix (dismissed):

Example/Sources/Swift/Info.plist
```diff
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
 	<key>CFBundleDisplayName</key>
-	<string>Checkout Example</string>
+	<string>Example</string>
```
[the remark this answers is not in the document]
siebeprojects: Android has two example apps. 

1. Shop Example
2. Checkout Example

Both SDKs should have the same naming, so also iOS should name it “Checkout Example” and not “Example”.
Later, we may add also a “Shop Example” to iOS SDK, so then also in iOS we will have “Shop Example” and “Checkout Example”.

CheckoutExample/Sources/Swift/Info.plist
```diff
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
 	<key>CFBundleDisplayName</key>
-	<string>Example</string>
+	<string>Checkout</string>
```
Sorix, line 8:
```suggestion
	<string>Checkout Example</string>
```

Sorix (changes requested):

Sorix (approved):

Sorix (approved):

siebeprojects (approved):

PR #159, opened 2021-12-01
[PCX-1004] Charge a PRESET account

Sorix:
**AS A** merchant developer
**I WANT** to have an API to charge a PRESET account
**SO THAT** I can do so from my summary page when the end-customer clicks the “Pay” button

### Acceptance criteria

1. The merchant calls a public charge method for the `presetAccount` which will trigger the `/charge` operation URL for the `presetAccount` object
1. The public method is consistent (to the extent possible) with the corresponding method on the Android SDK in naming, parameters, return values
1. Upon OPG response, a `PaymentResult` is returned to the merchant in exactly the same manner it would have been if charging any available network or saved method.
1. Example application should contain a draft button to make a charge request to test implemented feature.


Sources/Core/ChargePresetService/ChargePresetService.swift
```diff
@@ -0,0 +1,130 @@
...
+import UIKit
+import SafariServices
+
+@objc public class ChargePresetService: NSObject {
+    @objc public weak var delegate: ChargePresetDelegate?
+
+    private var redirectCallbackHandler: RedirectCallbackHandler?
+    private var paymentService: PaymentService?
+    private let connection: Connection
+    var presentedViewController: UIViewController?
```
caiomello, line 16: Does this need to be open to the outside?

Sources/Core/ChargePresetService/ChargePresetService.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    @objc public override convenience init() {
+        let connection = URLSessionConnection()
+        self.init(connection: connection)
+    }
+
+    internal init(connection: Connection) {
+        self.connection = connection
+        super.init()
+    }
```
caiomello, line 26:
```suggestion
    internal init(connection: Connection = URLSessionConnection()) {
        self.connection = connection
        super.init()
    }
```

Sources/ListResultProvider/ListResultProvider.swift
```diff
@@ -0,0 +1,118 @@
...
+// See the LICENSE file for more information.
+
+import Foundation
+
+class ListResultProvider {
+    private let paymentServicesFactory: PaymentServicesFactory
+    private let connection: Connection
+
+    /// Could contain list result after running `fetchListResult(from:completion:)`.
+    /// - Note: could containt list result even if `fetchListResult(from:completion:)` produced an error but list result was downloaded successfully.
```
caiomello, line 14:
```suggestion
    /// - Note: could contain list result even if `fetchListResult(from:completion:)` produced an error but list result was downloaded successfully.
```

Sources/UI/Service/PaymentSessionService/PaymentSessionServiceDelegate.swift
```diff
@@ -7,6 +7,6 @@
 import Foundation
 
 protocol PaymentSessionServiceDelegate: AnyObject {
-    func paymentSessionService(loadingDidCompleteWith result: Load<UIModel.PaymentSession, ErrorInfo>)
+    func paymentSessionService(loadingResultDidChange result: Load<UIModel.PaymentSession, ErrorInfo>)
```
caiomello, line 10: Did change? If this has a result, shouldn't it be "did finish"?

Sources/UI/Service/PaymentSessionService/PaymentSessionProvider.swift
```diff
@@ -100,82 +78,35 @@ class PaymentSessionProvider {
...
-        let filteredNetworks = SessionNetworks(
-            applicableNetworks: filteredPaymentNetworks,
-            accountRegistrations: filteredRegisteredNetworks,
-            presetAccount: listResult.presetAccount)
-
-        completion(filteredNetworks)
-    }
-
-    private func localize(filteredNetworks: SessionNetworks, completion: @escaping (Result<DownloadTranslationService.Translations, Error>) -> Void) {
+    private func fetchNetworksLocalizations(filteredNetworks: ListResultNetworks, completion: @escaping (Result<DownloadTranslationService.Translations, Error>) -> Void) {
```
caiomello, line 81:
```suggestion
    private func fetchLocalizations(for filteredNetworks: ListResultNetworks, completion: @escaping (Result<DownloadTranslationService.Translations, Error>) -> Void) {
```

Sources/Core/ChargePresetService/ChargePresetService.swift
Sorix: We need to have a public initializer, not the internal one.

Sources/Core/ChargePresetService/ChargePresetService.swift
Sorix: Did you mean `ChargePresetService`? If yes, it is used to make a charge request from the example app as a temporary solution without old public methods rework.

https://github.com/optile/checkout-ios/blob/34a599979021654aa37b73565f3299d662e7b544/ExampleCheckout/Sources/Swift/ViewController.swift#L49-L60

Sources/UI/Service/PaymentSessionService/PaymentSessionServiceDelegate.swift
Sorix: I changed it because the previous naming seems incorrect, it could have `loading` state that is not a final result.

https://github.com/optile/checkout-ios/blob/f867d09df861ed7829647a55bb9b4e88f6b01a2a/Sources/UI/Service/PaymentSessionService/PaymentSessionService.swift#L87-L91

Although my proposed naming was still incorrect so I made changes here: 10c84bc9add74705921b699254f02b88fdc66ad0 & f63195593b22df48ff26ec5e5fc9ffc28853e19c

Sources/Core/ChargePresetService/ChargePresetService.swift
caiomello: I was referring to `presentedViewController`.

Sources/Core/ChargePresetService/ChargePresetService.swift
caiomello: True, I missed the `public` keyword.

Sources/UI/Service/PaymentSessionService/PaymentSessionServiceDelegate.swift
Sorix: After our online discussion I separated logic in f2c6cd86dad766e229b0f623fd74a2ba28314a74

caiomello: Looks good!

Sources/Core/ChargePresetService/ChargePresetService.swift
Sorix: It is not public, I'm not clearly understand what do you mean.

caiomello: "open to the outside" = not `private`.

Sorix: Ah, that was kinda hard to understand :), changed in c3870bc7c55880f74c680977433a1d6710ec356c

Sources/UI/Service/PaymentSessionService/PaymentSessionProvider.swift
Sorix: I want to keep the name consistent with another method: `fetchSharedLocalization` & `fetchNetworksLocalizations`. `fetchLocalizations(for:)` could be misleading in that case.

Sources/Core/ChargePresetService/ChargePresetService.swift
caiomello: You're right, it made sense in my head but it wasn't that clear 😅

caiomello (approved):
🚀

Sources/ListResultProvider/ListResultProvider.swift
```diff
@@ -0,0 +1,118 @@
+// Copyright (c) 2021 Payoneer Germany GmbH
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Foundation
+
+class ListResultProvider {
```
siebeprojects, line 9: Should this be called: ListResultNetworksProvider, because it returns the list of filtered Networks right?

Sorix: No, it returns a list result and also filtered networks as well (as a helper).

https://github.com/optile/checkout-ios/blob/c3870bc7c55880f74c680977433a1d6710ec356c/Sources/ListResultProvider/ListResultNetworks.swift#L9-L18

siebeprojects (approved):

PR #160, opened 2021-12-05
[PCX-2344] Support charge preset account in the Example Checkout app

Sorix:
**AS** a developer
**I WANT** to test charging preset accounts 
**SO THAT** I know that charging preset account functionality works as expected

### Acceptance Criteria

1. The existing “Send Request” button is renamed to “Show Payment List”
1. A second button with the label "Charge Preset Account" is added beneath the “Show Payment List” button in main screen of the Checkout Example app
1. Clicking the “Charge Preset Account” adds a spinner (as is done for the “Pay” button in the normal CHARGE flow) and calls the `chargePresetAccount` method in the PaymentUI class
1. The operation result is displayed normally in the Example Checkout app, whether success or failure, also in the case where there is no `presetAccount` on the LIST object


ExampleCheckout/Sources/Swift/ViewController.swift
```diff
@@ -72,47 +125,25 @@ extension ViewController: ChargePresetDelegate {
...
+extension ViewController {
+    var tintColor: UIColor {
+        get { view.tintColor }
+        set {
+            view.tintColor = newValue
+            themeSwitch.onTintColor = newValue
+            textField.tintColor = newValue
+            showPaymentListButton.backgroundColor = newValue
+            chargePresetAccountButton.backgroundColor = newValue
+            setNavigationBarColor(to: newValue)
```
caiomello, line 142: Is there a reason for this to be an extension? Why not simply a private function in `ViewController`?

caiomello (approved):

ExampleCheckout/Sources/Swift/ViewController.swift
Sorix: It looked nicer for me and more Swifty. I'm okay with that change, changed in 5160a06b30054997eb24a1850456296368802387.

ExampleCheckout/Sources/Swift/ViewController.swift
```diff
@@ -131,16 +131,13 @@ extension ViewController: ChargePresetDelegate {
...
-        get { view.tintColor }
-        set {
-            view.tintColor = newValue
-            themeSwitch.onTintColor = newValue
-            textField.tintColor = newValue
-            showPaymentListButton.backgroundColor = newValue
-            chargePresetAccountButton.backgroundColor = newValue
-            setNavigationBarColor(to: newValue)
-        }
+    fileprivate func setTintColor(to color: UIColor) {
```
caiomello, line 134: `private` is enough, no?

```suggestion
    private func setTintColor(to color: UIColor) {
```

caiomello, line 134:
```suggestion
    private func setTintColor(to color: UIColor) {
```

ExampleCheckout/Sources/Swift/ViewController.swift
```diff
@@ -10,17 +10,36 @@ import PayoneerCheckout
...
     @IBOutlet weak var themeSwitch: UISwitch!
-    @IBOutlet weak var sendButton: UIButton!
+    @IBOutlet weak var showPaymentListButton: UIButton!
+    @IBOutlet weak var chargePresetAccountButton: ActivityIndicatableButton!
+
+    fileprivate var isEnabled: Bool = true {
+        didSet {
+            [showPaymentListButton, chargePresetAccountButton, textField, themeSwitch].forEach { $0?.isEnabled = isEnabled }
+
+            if isEnabled {
```
siebeprojects, line 20: Could you set a local variable like with value 1 or 0.6 and then have only once the following statement:
[showPaymentListButton, chargePresetAccountButton].forEach { $0?.backgroundColor = $0?.backgroundColor?.withAlphaComponent(1) }

Sorix: I can't because I apply 0.6 transparency for each background color for the specific view. E.g., if views have different background colors, that statement still will work and add transparency for each color.

Sorix: Great idea, I thought about that but I missed that while actually coding it :). Resolved in 3a6e42e.

caiomello (approved):

siebeprojects (approved):

PR #162, opened 2021-12-13
[PCX-2414] Write Preset flow automated UI tests

Sorix:
**AS A** developer
**I WANT** to avoid bugs for PRESET flow in future
**SO THAT** I should add automated tests for PRESET flow

### Acceptance Criteria
The following automated UI tests will be written:
1. Test charging the PresetAccount when the list does not have a preset account set
1. Test charging PresetAccount when the PresetAccount Object is from a credit card network
1. Test charging PresetAccount when the PresetAccount Object is from a redirect network (PayPal)
1. Test charging PresetAccount that requires 3DS2 validation using redirect (Magic number: 1,23)


ExampleCheckout/UITests/Abstract/NetworksTests.swift
```diff
@@ -9,46 +9,47 @@ import XCTest
...
 
-    /// Load an app and load networks list from list url.
-    func setupWithPaymentSession(transaction: Transaction) throws {
+    override func setUp() {
         continueAfterFailure = false
+        self.app = XCUIApplication()
+        app.launch()
 
-        self.app = try Self.setupWithPaymentSession(transaction: transaction)
+        super.setUp()
```
caiomello, line 17: `super` should be called at the beginning in `setUp` and at the end in `tearDown`. Did you add this here for a reason?

Sorix: I followed the same pattern as it required for overriding initializer, you need to init uninited properties and call `super.init()` after that.

Variable `app` acts like that property, but we use `setUp()` instead of `init` in `XCTestCase` in most cases, that's why I've chosen that way. It's not critical and if you want I could swap it.

caiomello: [Xcode's unit test template](https://www.hackingwithswift.com/example-code/testing/how-to-do-one-time-setup-for-your-tests) is explicit about where `super` should be called in `setUp()` and `tearDown()`. In theory this shouldn't cause any issues, but we have no guarantee from the documentation, so it's best if we are consistent with this.

Sorix: Resolved in 787dca7

caiomello (approved):

siebeprojects (approved):

PR #174, opened 2022-01-31
[PCX-2697] Update CI

Sorix:


.github/workflows/reusable_ui_tests.yml
```diff
@@ -1,44 +1,57 @@
...
+    # GitHub actions doesn't support arrays for arguments
+    inputs:
+      simulators:
+        default: '["iPhone 12 (14.5)"]'
+        description: 'Stringified JSON array listing devices'
+        required: false
+        type: string
+      schemes:
+        default: '["Swift"]'
+        description: 'Stringified JSON array listing XCode schemes'
```
caiomello, line 14:
```suggestion
        description: 'Stringified JSON array listing Xcode schemes'
```

.github/workflows/labeler.yml
```diff
@@ -1,11 +1,12 @@
-name: Apply labels
+name: Labeler
```
caiomello, line 1:
```suggestion
name: Labeling
```

.github/workflows/linter_tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Code styling
```
caiomello, line 1:
```suggestion
name: Linting
```

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,51 @@
+name: Integration tests
```
caiomello, line 1:
```suggestion
name: Pod integration test
```
(It only runs 1 test, right?)

.github/workflows/labeler.yml
Sorix: I used nouns for workflow names.

.github/workflows/integration_tests.yml
Sorix: Yes, but it could require renaming later if we add some more, category shouldn't contain info that's about CocoaPods. That's why I have a more specific description in a job.

.github/workflows/linter_tests.yml
Sorix: I used nouns for workflow names.

caiomello: Code styling and linting don't mean the same thing. This workflow doesn't _style_ the code, it only _lints_ it.

caiomello (approved):

.github/workflows/deploy_browserstack.yml
```diff
@@ -1,16 +1,17 @@
-name: Deploy to Browserstack
+name: Internal deployment
```
siebeprojects, line 1: The term deploy is incorrect, publish is better. 
Deploy is normally used to e.g. "deploy new features to the production server".

siebeprojects (dismissed):

siebeprojects (approved):

PR #198, opened 2022-05-02
[PCX-2999] Risk error reporting

caiomello:
- Removed Iovation risk provider from example app
- Centralized risk service instantiation in `Checkout` object
- Refactored risk implementation to work in a linear flow in order to store and send errors (see image below)
- Updated risk tests

image


Sources/Risk/RiskProviderError.swift
```diff
@@ -0,0 +1,55 @@
...
+            return "riskProviderInternalError"
+        case .externalFailure:
+            return "riskProviderExternalError"
+        }
+    }
+
+    public var reason: String {
+        switch self {
+        case .internalFailure(let reason, _, _), .externalFailure(let reason, _, _):
+            return String(reason.prefix(2000))
```
siebeprojects, line 25: 2000 could be defined as a static value like MAX_ERROR_LENGTH. Now it is not clear why it is chopped.

siebeprojects (dismissed):

Sources/Risk/RiskProviderError.swift
caiomello: https://github.com/optile/checkout-ios/pull/198/commits/e3c4906b206fbd547aa53c0339fc71ae91948958

siebeprojects (approved):

Sources/PayoneerCheckout/Risk/RiskService.swift
```diff
@@ -7,94 +7,129 @@
...
+                }
+            }
         }
     }
 
-    private func loadRiskProvider(using providerParameters: ProviderParameters) -> RiskProvider? {
-        guard let providerType = lookUp(providerCode: providerParameters.providerCode, providerType: providerParameters.providerType) else {
-            return nil
-        }
+    mutating func collectRiskData() -> [ProviderParameters] {
```
Sorix, line 55: It's not very clear why collecting risk data mutating `RiskService` itself, seems pretty strange. I feel that could lead to errors (if you collected risk data multiple times for example).

Sources/PayoneerCheckout/UI/Scenes/Input/Input.ViewController.PaymentController.swift
```diff
@@ -66,12 +65,12 @@ extension Input.ViewController.PaymentController {
...
 
 private struct PaymentRequestBuilder: Loggable {
-    let riskService: RiskService
+    var riskService: RiskService
 
     /// Create a payment request with data from `Input.Network`
     ///
     /// - Warning: extra elements section will be ignored (to be implemented)
-    func createPaymentRequest(for network: Input.Network) throws -> PaymentRequest {
+    mutating func createPaymentRequest(for network: Input.Network) throws -> PaymentRequest {
```
Sorix, line 73: It's not very clear why creating of payment request which should return `PaymentRequest` mutates to the whole builder class.

Could it lead to errors if you try to create a payment request multiple times for example?

Sources/Risk/RiskProviderError.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+// MARK: - Equatable
+
+extension RiskProviderError: Equatable {
+    public static func == (lhs: RiskProviderError, rhs: RiskProviderError) -> Bool {
+        switch (lhs, rhs) {
+        case (.internalFailure, .internalFailure), (.externalFailure, .externalFailure):
+            return true
```
Sorix, line 52: Is it equal? What if it has different internal failures (different codes for example)?

Sources/Risk/RiskProviderError.swift
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2022 Payoneer Germany GmbH
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Foundation
+
+private let errorStringMaximumLength = 2000
```
Sorix, line 9: Is it an OPG requirement? If yes, please add a comment about it.

Sorix (commented):
Could you fix UI tests failure, please?

ExampleCheckout/Sources/Objective-C/ViewController.m
```diff
@@ -28,7 +27,7 @@ - (void)viewDidLoad {
 - (IBAction)sendRequest:(UIButton *)sender {
     NSURL *url = [[NSURL alloc] initWithString:self.urlTextField.text];
     
-    CheckoutConfiguration *configuration = [[CheckoutConfiguration alloc] initWithListURL:url appearance:[CheckoutAppearance default] riskProviderClasses:@[[IovationRiskProvider class]] error:NULL];
```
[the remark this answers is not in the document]
caiomello: This was due to the submodules issue, after talking with @siebeprojects I removed it. We could keep it since compilation will fail even if we don't import/use it anywhere, I have no problem with reverting this.

This shouldn't have any impact on UI tests, risk has nothing to do with UI, does it?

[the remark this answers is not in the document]
siebeprojects: @Sorix no, removing risk provider from Example App will not have an impact on UI tests. Iovation Risk Provider is only to be used by Payoneer Checkout merchants and not for orchestration. So, it is better to leave it out.

Sources/IovationRiskProvider/IovationRiskProvider.swift
```diff
@@ -9,23 +9,33 @@ import Risk
...
         }
     }
 
     public func collectRiskData() throws -> [String: String?]? {
-        return ["blackbox": FraudForce.blackbox()]
+        let data = FraudForce.blackbox()
+
+        guard data.isEmpty == false else {
+            throw RiskProviderError.externalFailure(
+                reason: "Empty blackbox received from Iovation risk provider",
```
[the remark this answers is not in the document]
caiomello: Got this one from @siebeprojects I think. I agree with providing more info, but the way I understand it is that the message `An empty blackbox indicates there is a problem with the integration of the SDK or that the protection offered by the system may have been compromised.` refers to the integration between our iOS SDK and the Iovation library. It's irrelevant outside of that scope. Also, the Android Iovation library has no mention of empty blackboxes, and iOS and Android should match in this case.

Sources/PayoneerCheckout/Interface/CheckoutConfiguration.swift
```diff
@@ -38,7 +38,7 @@ enum CheckoutConfigurationError: Error {
     ///   - listURL: The URL contained in `links.self` on the response object from a create payment session request.
     ///   - appearance: The appearance settings to be used in the checkout UI. If not specified, a default appearance will be used.
     ///   - riskProviderClasses: An array of risk provider types.
-    @objc public convenience init(listURL: URL, appearance: CheckoutAppearance = .default, riskProviderClasses: [AnyClass] = []) throws {
+    @objc public convenience init(listURL: URL, appearance: CheckoutAppearance, riskProviderClasses: [AnyClass]) throws {
```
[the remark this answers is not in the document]
caiomello: Default arguments are not supported in Objective-C.

Sources/PayoneerCheckout/Model/ProviderParameters.swift
```diff
@@ -16,9 +16,22 @@ public class ProviderParameters: NSObject, Codable {
     /// An array of parameters
     let parameters: [Parameter]?
 
-    init(providerCode: String, providerType: String?, parameters: [Parameter]?) {
+    init(providerCode: String, providerType: String?, parameters: [Parameter]) {
```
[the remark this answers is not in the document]
caiomello: Optional sequences are kind of an anti-pattern because you end up with two ways of representing an empty value. In this object the value comes from the backend so it's out of our control, but this custom initializer is ours and it can be a little bit better :)

Sources/PayoneerCheckout/Risk/RiskService.swift
caiomello: Good point! I could store the data collection errors in a local array instead and loop through them after looping through the stored errors from `loadRiskProviders`. It makes the function a bit more complex, but it will avoid sending duplicate data collection errors.

Sources/Risk/RiskProviderError.swift
caiomello: It's defined in the ticket. @siebeprojects is there any documentation I can refer to here?

Sources/PayoneerCheckout/UI/Scenes/Input/Input.ViewController.PaymentController.swift
caiomello: https://github.com/optile/checkout-ios/pull/198/commits/d57f9a81ab60d1cc2411088d59418a2c3519ecc2

Sources/PayoneerCheckout/Risk/RiskService.swift
caiomello: https://github.com/optile/checkout-ios/pull/198/commits/d57f9a81ab60d1cc2411088d59418a2c3519ecc2

Sources/Risk/RiskProviderError.swift
caiomello: The definition of equal is up to us. I use this for the tests and since we don't perform any logic on these errors this definition is reasonable right now. If this changed, then the definition of equal here and the tests would need to be updated.

Sources/IovationRiskProvider/IovationRiskProvider.swift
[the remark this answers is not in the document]
siebeprojects: The Iovation Android SDK does not define that an empty blackbox can be returned, instead we expect the the Iovation SDK to throw an exception when something goes wrong. Thats why we are not handling the empty blackbox.
We have 2000 characters for the error message, so we could send the longer text which is more descriptive.

Sources/Risk/RiskProviderError.swift
siebeprojects: The requirement is is that we should not send an error message longer than 2000 characters. I do not believe we need to refer to any documentation about that from code.

Sorix: I still think we should provide a link because in the future we will forget about this ticket and it will look like a "magic number".

```suggestion
// Defined in https://optile.atlassian.net/browse/PCX-2999
private let errorStringMaximumLength = 2000
```

Sources/Risk/RiskProviderError.swift
Sorix: It will be easy to miss such things in the future, we should use a custom method or make equatable more precise.

If it is used only for testing, is it possible to move this extension to the `Tests` module? In `Tests` we could ignore such things.

caiomello: https://github.com/optile/checkout-ios/pull/198/commits/b6302a344c9693e824317b6ab3a17beaf0fedbd1

siebeprojects (approved):

PR #199, opened 2022-05-02
[PCX-3100] Support redirect

Sorix:


Sources/Redirection/RedirectionError.swift
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022 Payoneer Germany GmbH
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Foundation
+
+enum RedirectionError: LocalizedError {
```
siebeprojects, line 9: RedirectError is a better name.

Package.swift
```diff
@@ -29,9 +29,10 @@ let package = Package(
 
         // Payment Services
         .target(name: "Payment", dependencies: ["Networking"]),
+        .target(name: "Redirection", dependencies: ["Networking"]),
         .target(
             name: "BasicPaymentService",
-            dependencies: ["Networking", "Payment"],
+            dependencies: ["Networking", "Payment", "Redirection"],
```
siebeprojects, line 35: I believe "Redirect" is a better name than "Redirection".

Sorix: @caiomello what do you think? I'm okay with `Redirect`, I used the same pattern as we used for `Networking`, `Logging`.

siebeprojects: @caiomello I prefer "redirect" similar as "payment", those names work for e.g. networking and logging, but  I don't think "redirecting" is better either.

Sources/PayoneerCheckout/UI/Scenes/Input/Table/Input.Table.Controller.swift
```diff
@@ -95,19 +95,15 @@ extension Input.Table {
             // I disable animation for iOS 12 and lower because it cause animation bugs (it's related to dynamic cell size calculations) and animation for that block is not important.
             // Radar: http://www.openradar.me/23728611
             // Article describing the same situation: https://jakubturek.com/uicollectionview-self-sizing-cells-animation/
-            if #available(iOS 13, *) {
-                // Everything is okay, nothing to disable
-            } else {
+            if #unavailable(iOS 13) {
```
caiomello, line 98: Nice, didn't know about this inverted check, looks like it was introduced in 5.6.

Sources/Redirection/RedirectCallbackHandler.swift
```diff
@@ -0,0 +1,83 @@
...
+    var asDictionary: [String: String] {
+        var dict = [String: String]()
+        for queryItem in self {
+            guard let value = queryItem.value else { continue }
+            dict[queryItem.name] = value
+        }
+
+        return dict
+    }
+}
```
caiomello, line 76: `URLQueryItem` is already a key-value pair, so a `[URLQueryItem]` is essentially already a `[String: String]` and this transformation is redundant, no?

Sources/Redirection/RedirectCallbackHandler.swift
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2022 Payoneer Germany GmbH
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Foundation
+import Networking
+
+class RedirectCallbackHandler {
```
caiomello, line 10:
```suggestion
final class RedirectCallbackHandler {
```

Sources/Redirection/RedirectCallbackHandler.swift
```diff
@@ -0,0 +1,83 @@
...
+        removeObserver()
+
+        self.completionBlock = completion
+
+        // Received payment result notification
+        receivePaymentNotificationToken = NotificationCenter.default.addObserver(forName: openAppWithURLNotificationName, object: nil, queue: .main) { [removeObserver] notification in
+            guard let url = notification.userInfo?["url"] as? URL else { return }
+            self.handle(receivedURL: url)
+            removeObserver()
+        }
```
caiomello, line 30: Both `[removeObserver]` and `self` in line 28 are a strong captures. Best to use `[weak self]` instead:

```suggestion
        receivePaymentNotificationToken = NotificationCenter.default.addObserver(forName: openAppWithURLNotificationName, object: nil, queue: .main) { [weak self] notification in
            guard let url = notification.userInfo?["url"] as? URL else { return }
            self?.handle(receivedURL: url)
            self?.removeObserver()
        }
```

Sources/Redirection/RedirectCallbackHandler.swift
```diff
@@ -0,0 +1,83 @@
...
+        // Remove observers to avoid duplicate completion calls if `addObservers()` were accidentally called twice
+        removeObserver()
+
+        self.completionBlock = completion
+
+        // Received payment result notification
+        receivePaymentNotificationToken = NotificationCenter.default.addObserver(forName: openAppWithURLNotificationName, object: nil, queue: .main) { [removeObserver] notification in
+            guard let url = notification.userInfo?["url"] as? URL else { return }
+            self.handle(receivedURL: url)
+            removeObserver()
```
caiomello, line 29: This flow is a bit confusing to me... observers are automatically removed during `deinit`, so removing it in line 21 before adding a new observer should be enough, no? Removing it here will make it so that this only gets triggered once, but still you can always call `addObserver` again, which defeats the purpose...

I think you could probably make this simpler by using a function instead of a block-based observer.

Package.swift
caiomello: This is a tricky one so the confusion is understandable, but `Redirect` is more appropriate for sure since technically we perform a "redirect", not a "redirection". But I'm not sure why this needs to be a separate module... I added a comment here https://github.com/optile/checkout-ios/pull/199#discussion_r863587054.

Sources/PayoneerCheckout/Extensions/NSNotification+didReceiveCallbackFromURL.swift
```diff
@@ -0,0 +1,11 @@
...
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Foundation
+
+extension NSNotification.Name {
+    static let didReceiveCallbackFromURL = NSNotification.Name(rawValue: "PayoneerCheckoutDidReceiveCallbackFromURL")
+}
```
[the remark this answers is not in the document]
Sorix: I used that because Apple extends Notifications for each module that sends notifications, I don't see the real problem here (for notifications especially), I will recheck it.

Sources/Redirection/RedirectCallbackHandler.swift
```diff
@@ -0,0 +1,83 @@
...
+
+        return dict
+    }
+}
+
+// MARK: - Constants
+
+private extension String {
+    static var interactionCodeKey: String { "interactionCode" }
+    static var interactionReasonKey: String { "interactionReason" }
```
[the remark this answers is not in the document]
Sorix: I agree that we could move out it from `String` extension but I don't want to mark it as `static let` (it will occupy memory for a whole lifecycle). That's why I used computed variables.

What do you think about that idea 4203a392a96c05dbe9ce2016e2fbeff561e1d0bc?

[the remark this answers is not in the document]
Sorix: I agree that we could move out it from `String` extension but I don't want to mark it as `static let` (it will occupy memory for a whole lifecycle). That's why I used computed variables.

What do you think about that idea 4203a392a96c05dbe9ce2016e2fbeff561e1d0bc?

ExampleCheckout/Sources/Swift/AppDelegate.swift
```diff
@@ -5,14 +5,22 @@
...
+            Checkout.handle(callbackURL: url)
+        }
+
         return true
     }
+
+    private func isPayoneerCheckout(callbackURL url: URL) -> Bool {
+        let payoneerCheckoutScheme = "com.payoneer.checkout.examplecheckout.mobileredirect"
+        return url.scheme == payoneerCheckoutScheme
+    }
```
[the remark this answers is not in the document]
Sorix: I implemented two different solutions in this file.

#### Checking URL scheme
https://github.com/optile/checkout-ios/blob/8aeb43329b73abb70d2a7d085aa5d0c81619c589/ExampleCheckout/Sources/Swift/AppDelegate.swift#L22-L25
I didn't do it in the past, but it should be done before. The only merchant knows what scheme will be used in a redirect URL (now it's based on an application ID that we shouldn't have access to, a merchant could use any other option in theory).

So it doesn't matter if the merchant should send a notification and call a method; the merchant should check if that URL is related to our framework.

#### Notification vs static method
https://github.com/optile/checkout-ios/blob/8aeb43329b73abb70d2a7d085aa5d0c81619c589/ExampleCheckout/Sources/Swift/AppDelegate.swift#L16
I think it's easier for a merchant to use a static method than notifications. For notifications, merchants should know the notification's name and pass data via `userInfo` without strict formatting. It works like giving data with unstructured JSON. That's why I created such a "proxy" object.

Sources/Redirection/RedirectCallbackHandler.swift
Sorix: I think I made this decision because `URLQueryItem` doesn't support subscript by key and doesn't have `removeValue(forKey:)` method. It doesn't look like it is a `Dictionary`.

Sources/Redirection/RedirectCallbackHandler.swift
Sorix: 1. The observer could be removed externally: https://github.com/optile/checkout-ios/blob/8aeb43329b73abb70d2a7d085aa5d0c81619c589/Sources/Redirection/RedirectController.swift#L42-L46
2. If I don't remove the observer manually it doesn't allow `self` to deallocate (I think because Notification creates a strong reference to `self`). I just checked it, `deinit` is never called in that case. So anyway I should remove that observer manually.

---

I wrote an example code, did you mean that (replace the whole file)?

```swift
import Foundation
import Networking

final class RedirectCallbackHandler {
    private let constants = Constant()

    private let openAppWithURLNotificationName: NSNotification.Name
    private var completionBlock: ((Result<OperationResult, Error>) -> Void)?
    private var receivePaymentNotificationToken: NSObjectProtocol?

    init(openAppWithURLNotificationName: NSNotification.Name) {
        self.openAppWithURLNotificationName = openAppWithURLNotificationName
    }

    func addObserver(completion: ((Result<OperationResult, Error>) -> Void)?) {
        // Remove observers to avoid duplicate completion calls if `addObservers()` were accidentally called twice
        removeObserver()

        self.completionBlock = completion

        // Received payment result notification
//        receivePaymentNotificationToken = NotificationCenter.default.addObserver(forName: openAppWithURLNotificationName, object: nil, queue: .main) { [removeObserver] notification in
//            guard let url = notification.userInfo?["url"] as? URL else { return }
//            self.handle(receivedURL: url)
//            removeObserver()
//        }
        NotificationCenter.default.addObserver(self, selector: #selector(didReceiveNotification), name: openAppWithURLNotificationName, object: nil)
    }

    func removeObserver() {
        if let receivePaymentNotificationToken = receivePaymentNotificationToken {
            NotificationCenter.default.removeObserver(receivePaymentNotificationToken)
            self.receivePaymentNotificationToken = nil
        }

        self.completionBlock = nil
    }

    @objc private func didReceiveNotification(_ notification: Notification) {
        print("OBSERVER CALLED")
        guard let url = notification.userInfo?["url"] as? URL else { return }
        handle(receivedURL: url)
    }

    private func handle(receivedURL: URL) {
        guard
            let components = URLComponents(url: receivedURL, resolvingAgainstBaseURL: false),
            var queryItems = components.queryItems?.asDictionary,
            let interactionCode = queryItems[constants.interactionCodeKey],
            let interactionReason = queryItems[constants.interactionReasonKey]
        else {
            completionBlock?(.failure(RedirectionError.missingOperationResult))
            return
        }

        queryItems.removeValue(forKey: constants.interactionCodeKey)
        queryItems.removeValue(forKey: constants.interactionReasonKey)

        let interaction = Interaction(code: interactionCode, reason: interactionReason)
        let parameters: [Parameter] = queryItems.map { .init(name: $0.key, value: $0.value) }
        let redirect = Redirect(url: receivedURL, method: .GET, parameters: parameters)

        let operationResult = OperationResult(resultInfo: "OperationResult received from the mobile-redirect webapp", links: nil, interaction: interaction, redirect: redirect)

        completionBlock?(.success(operationResult))
    }

    deinit {
        print("DEINIT HAPPENED")
    }
}

private extension Sequence where Element == URLQueryItem {
    var asDictionary: [String: String] {
        var dict = [String: String]()
        for queryItem in self {
            guard let value = queryItem.value else { continue }
            dict[queryItem.name] = value
        }

        return dict
    }
}

// MARK: - Constants

private struct Constant {
    let interactionCodeKey = "interactionCode"
    let interactionReasonKey = "interactionReason"
}
```

Sources/PaymentServices/BasicPaymentService/BasicPaymentService.swift
```diff
@@ -4,13 +4,18 @@
 // This file is open source and available under the MIT license.
 // See the LICENSE file for more information.
 
-import Foundation
+import UIKit
 import Networking
 import Payment
-import UIKit
+import Redirection
```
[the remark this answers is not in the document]
Sorix: Not every network could have redirect (Apple Pay for example). 

I will merge it because supporting different service types will be a more complex task. We shouldn't limit ourselves to `PaymentService` type in the future, we should use multiple protocols and `PaymentController` should know how to deal with multiple protocols, redirect is one example because it requires information from AppDelegate.

 Merged in 7b0d1b76678735a7f3cc75315075eea2a2ac8361

Package.swift
Sorix: ~Renamed in f4a458264808cfceea3ba88b80c67e48185aecb7~
Removed `Redirect` module in 7b0d1b76678735a7f3cc75315075eea2a2ac8361

Sources/Redirection/RedirectionError.swift
Sorix: Renamed in 63311bf197b5e3472c05be4183afd95de9faaa6e

Sources/Redirection/RedirectURLBuilder.swift
```diff
@@ -0,0 +1,60 @@
...
+            throw BuilderError(errorDescription: "Unable to build redirect URL from components")
+        }
+
+        return url
+    }
+}
+
+private struct BuilderError: LocalizedError {
+    var errorDescription: String?
+}
```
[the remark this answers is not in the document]
Sorix: I renamed in 7d5f96b836198b5cceed6b8455003b5587f5ff79.

The position of such small structures is a question to discuss. Sometimes it's better to have it in the header but when you have multiple "small" structures or they're more complex than just 2-3 lines it doesn't look good.

I prefer to have the main class or structure at the top but that's discussable. Let's have an agreement about this in an offline discussion.

Sources/Redirection/RedirectCallbackHandler.swift
Sorix: Changed in 1c579211e5b4eac73ec414e77dc699ee1e6eec7a

Sources/Redirection/RedirectionError.swift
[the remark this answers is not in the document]
caiomello: I agree `LocalizedError` is the way to go, I just want to be careful with inconsistencies, otherwise we end up with some errors that conform to `Error` and others that conform to `LocalizedError`. We can keep it like this since we have a future task to look at error handling in general.

ExampleCheckout/Sources/Swift/AppDelegate.swift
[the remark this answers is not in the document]
caiomello: I'm really not sure about this and since this is a breaking change I feel it's probably best if we do it separately. What do you think?

Sources/PayoneerCheckout/Extensions/NSNotification+didReceiveCallbackFromURL.swift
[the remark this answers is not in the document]
caiomello: https://github.com/optile/checkout-ios/pull/199#discussion_r866652319

caiomello (approved):

siebeprojects (approved):

PR #200, opened 2022-05-05
[PCX-3260] Iovation risk provider fixes

caiomello:
- Restore missing comma that caused compilation to fail
- Updated risk provider error keys
- Added a simple test to make sure Iovation risk provider can be compiled


Sources/Risk/RiskProviderError.swift
```diff
@@ -16,9 +16,9 @@ public enum RiskProviderError: Error, Equatable {
     public var name: String {
         switch self {
         case .internalFailure:
-            return "riskProviderInternalError"
+            return "riskPluginInternalError"
```
Sorix, line 19: Should we rename providers as well? 🙂

Sorix (approved):

siebeprojects (approved):

PR #201, opened 2022-05-05
[PCX-3273] Update Xcode version on CI

Sorix:
* UI tests for a single device that we use when merging to `develop` will use the latest Xcode version instead of the previous one. It is required to unblock compilation for the latest pull request.
* Seems that GitHub resolved a bug https://github.com/github/feedback/discussions/10679. I could revert UI tests to run only on the pull requests.
* Keep Gemfile up to date (that could resolve security issues in _fastlane_ as well).

More information in self-review.

### Pre-merge requirements
- [ ] Update merge rule requirement from `Single device / Swift, iPhone 12 (14.5)` to `Single device / Swift, iPhone 13 (15.2)` (@siebeprojects)


.github/workflows/ui_tests.yml
```diff
@@ -17,8 +12,7 @@ jobs:
     name: Single device
     uses: ./.github/workflows/reusable_ui_tests.yml
     with:
-      install-simulators-command: >-
-        sudo ln -s $XCODE_12_DEVELOPER_DIR/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime /Library/Developer/CoreSimulator/Profiles/Runtimes/iOS\ 14.5.simruntime
```
Sorix, line 21: Without that line, the latest Xcode will be loaded for single device tests.

.github/workflows/ui_tests.yml
```diff
@@ -17,8 +12,7 @@ jobs:
...
+      simulators: '["iPhone 13 (15.2)"]'
```
Sorix, line 15: Now tests will be run against iOS 15.2 instead of iOS 14.5. We used iOS 14.5 before because of a slow typing bug that looks like Apple fixed it.

.github/workflows/ui_tests.yml
```diff
@@ -1,11 +1,6 @@
 name: UI tests
 
-on:
-  # I use push instead of pull request because of bug with reusable workflows.
-  # You can't use reusable workflows when action was triggered by pull request.
-  # Discussion: https://github.com/github/feedback/discussions/10679
-  push:
-    branches-ignore: master
+on: pull_request
```
Sorix, line 4: Seems that GitHub resolved a bug https://github.com/github/feedback/discussions/10679, I could revert back UI tests to run only on the pull requests.

Sorix:
@siebeprojects please update test requirement from `Single device / Swift, iPhone 12 (14.5)` to `Single device / Swift, iPhone 13 (15.2)` before merge.

.github/workflows/ui_tests.yml
caiomello: In that case we should move all tests to 15.2, no?

Also we can remove the paste workaround in UI tests, but not sure if we should do it in this ticket.

siebeprojects (approved):

.github/workflows/ui_tests.yml
caiomello: > We should test when we merge to master on last 3 supported platforms at least

I meant when merging to `develop`. It should test against the most used version, which is 15 now, but yeah let's do it separately.

caiomello (approved):

.github/workflows/ui_tests.yml
Sorix: In that PR I already did it, now tests will be run against `15.2`

PR #203, opened 2022-05-18
[PCX-3332] Remove outdated versions on Browserstack

Sorix:
Only one version will be presented on Browserstack per each branch. When the new version is uploaded, the previous one will be replaced.

**Browserstack API**: https://www.browserstack.com/app-live/rest-api

#### Changelog
* Create a script to delete previous versions.
   ```bash
   python .github/scripts/browserstack_delete.py <custom_id>
   ```
* Generate custom ID (`iOS_branch_name`) and binary name.
* 🆕 Delete previous versions before uploading the new one.
* 🆕 Binaries will be uploaded for each pull request (previously only for labeled ones).

### DIFF
--- a/.github/scripts/browserstack_delete.py
+++ b/.github/scripts/browserstack_delete.py
@@ -26,8 +26,7 @@ def delete_recent_apps(custom_id: str):
     print(f'Deleting binaries with custom_id {custom_id}')
 
     recent_uploads = get_recent_uploads(custom_id)
-
-    if recent_uploads["message"] is not None:
+    if isinstance(recent_uploads, dict) and recent_uploads["message"] is not None:
         if recent_uploads["message"] == "No results found":
             print("Nothing to delete")
             return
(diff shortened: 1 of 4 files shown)


caiomello (dismissed):

Sorix:
@siebeprojects I didn't make prefix `Example_` for application because now we have only one application for iOS. When we create the second one we will add this prefix, and now for clarity and shorter naming we could strip it.

siebeprojects (dismissed):

caiomello (approved):

.github/scripts/browserstack_delete.py
```diff
@@ -26,8 +26,7 @@ def delete_recent_apps(custom_id: str):
     print(f'Deleting binaries with custom_id {custom_id}')
 
     recent_uploads = get_recent_uploads(custom_id)
-
-    if recent_uploads["message"] is not None:
+    if isinstance(recent_uploads, dict) and recent_uploads["message"] is not None:
```
siebeprojects, line 29: Would it be better to check if the returned JSON Object is off type array [] and has elements instead of checking that there is a message with a certain value?
I don't understand the construction of having two nested if statements.

siebeprojects (approved):

PR #219, opened 2022-06-13
[PCX-3324] Integrate Braintree SDK

Sorix:
The Braintree SDK needs to be imported into our SDK and used to generate a PKPaymentRequest to inject into the Apple Pay framework.

* Create the Braintree Apple Pay payment service module.
* Make `onSelect` call.
* Create a Braintree API client.
* Be able to generate a `PKPaymentRequest` object using the Braintree SDK.


Sources/Networking/Requests/NetworkRequest.Operation.swift
```diff
@@ -14,15 +14,22 @@ public extension NetworkRequest {
...
         public typealias Response = OperationResult
 
         /// - Parameter url: value from `links.operation` for charge operation
-        public init(from url: URL, body: Body) {
+        public init(from url: URL, account: [String: String]?, autoRegistration: Bool?, allowRecurrence: Bool?, providerRequest: ProviderParameters?, providerRequests: [ProviderParameters]?) {
             self.url = url
-            self.chargeBody = body
+
+            self.body = OperationData(
+                account: account,
```
siebeprojects, line 26: _Not a blocker_
Instead of having a map of strings, you could have a model class AccountInputData that contains a map or the fixed parameters as in Android.

Sources/PaymentServices/ApplePayBraintreePaymentService/BraintreeClientFabric.swift
```diff
@@ -0,0 +1,23 @@
...
+// https://www.payoneer.com
+//
+// This file is open source and available under the MIT license.
+// See the LICENSE file for more information.
+
+import Payment
+import Networking
+import BraintreeApplePay
+
+struct BraintreeClientFabric {
```
siebeprojects, line 11: The name "Fabric" is strange here. Does this come from Braintree or is that from us?

Sources/Networking/Requests/NetworkRequest.Operation.swift
Sorix: This will be outside of the scope of this ticket and require other implementation.

Sources/PaymentServices/ApplePayBraintreePaymentService/BraintreeClientFabric.swift
Sorix: Yes, it's a better naming, thanks. Changed in 35b4222e8a4c1190a1305c4c014c3ef3f82d035f

Sources/Networking/Model/OperationData.swift
```diff
@@ -0,0 +1,28 @@
...
+    /// Provider request parameters which should be provided to the target payment provider adapter to complete the operation.
+    public var providerRequest: ProviderParameters?
+
+    /// List of provider request parameters which should be provided to the target payment provider adapter to complete the operation.
+    public var providerRequests: [ProviderParameters]?
+
+    /// Customer web browser data.
+    internal var browserData: BrowserData?
+
+    // Not used variables: checkboxes: Object, style: ClientStyle
```
Fbada006, line 27: Why is this comment here?

Sources/PaymentServices/ApplePayBraintreePaymentService/ApplePayBraintreePaymentService.swift
```diff
@@ -0,0 +1,110 @@
...
+    let redirectController: RedirectController
+    let connection: Connection
+
+    public init(connection: Connection, openAppWithURLNotificationName: NSNotification.Name) {
+        self.connection = connection
+        self.redirectController = RedirectController(openAppWithURLNotificationName: openAppWithURLNotificationName)
+    }
+
+    public static func isSupported(networkCode: String, paymentMethod: String?) -> Bool {
+        return networkCode == "APPLEPAY" && paymentMethod == "WALLET"
```
Fbada006, line 22: Can you not set these as constants?

Sources/PaymentServices/ApplePayBraintreePaymentService/ApplePayBraintreePaymentService.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // MARK: Process payment
+
+    public func processPayment(operationRequest: OperationRequest, completion: @escaping PaymentService.CompletionBlock, presentationRequest: @escaping PaymentService.PresentationBlock) {
+        // OnSelect
```
Fbada006, line 28: This comment here does not really communicate anything, does it?

Sources/PaymentServices/ApplePayBraintreePaymentService/PaymentRequestBuilder.swift
```diff
@@ -0,0 +1,61 @@
...
+                completion(.failure(error))
+            }
+        }
+    }
+
+    /// Add missing data to `PKPaymentRequest`
+    /// - Important: modifies `PKPaymentRequest`
+    private func complement(paymentRequest: PKPaymentRequest) throws {
+        // Overwrite properties filled by Braintree if they're present in OperationResult
+        if let appleMerchantId = providerResponse.parameters?["appleMerchantId"] {
```
Fbada006, line 43: Same with these params. Can you use constants?

Sources/Networking/Model/OperationData.swift
Sorix: I placed this reminder because OPG has these properties in `OperationData` model but we don't use them in the code. I constructed this model from Swagger's API file that's why I left this comment not to forget in the future (all models in `Model/` folder are in from OPG).

```
OperationData:
type: object
properties:
  account:
    $ref: '#/definitions/AccountInputData'
    description: Account data submitted by a customer from account-form; presence of account or other fields depends on a selected payment network
  autoRegistration:
    type: boolean
    description: If set to `true` the account will be registered for further payments
  allowRecurrence:
    type: boolean
    description: If set to `true` the account will be registered for further recurring payments
  checkboxes:
    type: object
    description: |
        Map of selected checkboxes; name reflects the checkbox distinguish name, value reflects its
        state - `true` for selected, `false` for unchecked
    additionalProperties:
      type: boolean
  providerRequest:
    $ref: '#/definitions/ProviderParameters'
    description: |
        Provider request parameters which should be provided to the target payment provider adapter to complete the operation
  providerRequests:
    type: array
    description: |
        List of provider request parameters which should be provided to the target payment provider adapter to complete the operation.
    items:
      $ref: '#/definitions/ProviderParameters'
  browserData:
    $ref: '#/definitions/BrowserData'
    description: Customer web browser data.
  style:
    $ref: '#/definitions/ClientStyle'
    description: Allows to style payment page
```

siebeprojects (approved):

PR #230, opened 2022-08-01
[PCX-3760] Add option to manually upload to Browserstack

Sorix:
> https://optile.atlassian.net/browse/PCX-3760

## Overview
Add option to manually upload to Browserstack through GitHub Actions.

## Checklist
_Not applicable_

### DIFF
--- a/.github/workflows/publish_browserstack.yml
+++ b/.github/workflows/publish_browserstack.yml
@@ -1,6 +1,8 @@
 name: Publish to BrowserStack
 
 on:
+  workflow_call:
+
   pull_request:
     types:
       - opened
--- a/.github/workflows/publish_browserstack.yml
+++ b/.github/workflows/publish_browserstack.yml
@@ -1,7 +1,7 @@
 name: 'ACTION: Publish to BrowserStack'
 
 on:
-  workflow_call:
+  workflow_dispatch:
 
   pull_request:
     types:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_browserstack.yml
```diff
@@ -1,6 +1,8 @@
 name: Publish to BrowserStack
 
 on:
+  workflow_call:
```
caiomello, line 4: It's supposed to be `workflow_dispatch`, no?

Sorix: Thank you, changed in cd3671b7fc4626a5834315c75ed2bfc935e9d8f6

.github/workflows/publish_browserstack.yml
```diff
@@ -1,6 +1,8 @@
-name: Publish to BrowserStack
+name: 'ACTION: Publish to BrowserStack'
```
caiomello, line 1: Do we really need to use a separate workflow and name?

caiomello (approved):

.github/workflows/publish_browserstack.yml
Sorix: Yes, I follow a pattern in other repositories we have, to have a clear indication for Tal for example that this action could be run manually. I don't like how it looks when it was run automatically, but I don't see a better solution.

PR #237, opened 2022-08-17
[PCX-3918] Apple Pay UI tests failing

Sorix:
> https://optile.atlassian.net/browse/PCX-3918

## Overview

Apple Pay tests are failing on iOS 13 devices. More information in GitHub Actions run log: https://github.com/optile/checkout-ios/runs/7876854263?check_suite_focus=true.

#### Bug reasons
1. We couldn't interact with Apple Pay in iOS 14 or lower.
2. Apple Pay view controller for iOS14 differs from iOS15.

#### Changes
* Tests are now skipped for iOS 14- because we couldn't really test it.
* We shouldn't rely on a full string to detect labels, the new method `containing` was introduced to avoid it (it could help support multilanguage tests in the future). The extension is marked as `private` because it's used now only in a single class, if it will be used in other places it could be moved to a global extension.

_Tests could be skipped using Test Suites but it is out of the scope of this ticket and will have a niche usage in our framework._

### DIFF
--- a/ExampleCheckout/UITests/ApplePayBraintreeTests.swift
+++ b/ExampleCheckout/UITests/ApplePayBraintreeTests.swift
@@ -120,8 +120,8 @@ private extension ApplePayBraintreeTests {
             throw "Timeout: waiting of Apple Pay presentation"
         }
 
-        // iOS 13 doesn't allow interaction with Apple Pay view controller
-        // iOS 14+ use a button instead of cell so we could guess version using a code below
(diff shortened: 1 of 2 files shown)


Sorix:
**Don't merge**. The base is set to `master` to trigger multiple tests on multiple devices.

Sorix:
Tests against `master` with multiple iOS versions passed.

siebeprojects (approved):

siebeprojects (approved):

Fbada006 (approved):

